PR #12, opened 2017-10-05
Allow EDDTableFromErddap to add/edit global attributes

kwilcox:


BobSimons:
Sorry, I'm going to decline your pull request.

There is a reason why EDDTableFromErddap and EDDGridFromErddap don't allow a global <addAttributes> section in datasets.xml for that dataset:
1) Let's say it was allowed and you added a new attribute.
2) When a user asks your ERDDAP for the global attributes, the new attribute will appear.
3) But when a user asks your ERDDAP for a data file, your ERDDAP redirects the request to the source ERDDAP. That ERDDAP is unaware of the new attribute. So if it creates a data file with metadata, e.g., a .nc file, the metadata won't have the new attribute.

So allowing <addAttributes> for EDD...FromErddap datasets would lead to inconsistencies. 
So it is not allowed.

There are two work-arounds:

1) Convince the admin of the source ERDDAP to make the changes that you want to the metadata.

2) Instead of EDDTableFromErddap, use EDDTableFromDapSequence.
Or instead of EDDGridFromErddap, use EDDGridFromDap.
Those EDD types allow you to connect efficiently to a dataset on a remote ERDDAP (but without redirecting data requests) and they allow you to include <addAttributes> section in datasets.xml. 
One other difference: you will need to manually subscribe to the remote dataset, so that the dataset on your ERDDAP will be notified (via the flagUrl) when there are changes to the remote dataset.
Thus, you are creating a new dataset, instead of linking to a remote dataset. 

I will improve the ERDDAP documentation related to this.

I hope that helps.

shane-axiom:
Would you consider allowing `addAttributes` in `EDD*FromErddap` when `redirect` is false? I understand the concerns when data requests are redirected, but there is great utility in being able to massage attributes to present a dataset in different contexts.

BobSimons:
I posted a solution in my Oct 10, 2017 comment above:

"
Instead of EDDTableFromErddap, use EDDTableFromDapSequence.
Or instead of EDDGridFromErddap, use EDDGridFromDap.
Those EDD types allow you to connect efficiently to a dataset on a remote ERDDAP (but without redirecting data requests) and they allow you to include section in datasets.xml.
One other difference: you will need to manually subscribe to the remote dataset, so that the dataset on your ERDDAP will be notified (via the flagUrl) when there are changes to the remote dataset.
Thus, you are creating a new dataset, instead of linking to a remote dataset.
"

This approach lets you modify global and variable attributes, just like other dataset types in ERDDAP.
Why is this approach not sufficient?

shane-axiom:
`EDDTableFromErddap` is far simpler and more maintainable than `EDDTableFromDapSequence` since it doesn't require a `dataVariables` declaration. `EDDTableFromDapSequence` has to be updated when the source ERDDAP dataset's variable change.

BobSimons:
1. I'm not going to change EDDTableFromErddap. It fills a need with great simplicity and it functions correctly.
2. If I were to make a variant of EDDTableFromErddap that allows admins to change the metadata and other things, it would be 99% identical to EDDTableFromDapSequence. If you want to be able to change some things, then you have to accept all the boilerplate that comes with that system (e.g., <dataVariable>s). [Note that some people want to change some things, and other people want to change other things. I'm not going to make a variant just for the things you currently want to change.] So I see no point in making that new variant. Just use EDDTableFromDapSequence.
3. I don't see why making <dataVariable> declarations is a big deal (or a small deal). If you use generateDatasetsXml with the EDDTableFromDapSequence options, it will generate a chunk of xml which exactly mimics the source dataset. It will have all the <dataVariable> tags. All you then have to do is edit the attributes, variable <destinationName>s or whatever you want to edit.

shane-axiom:
That's what I expected but I had to ask! Duplicating config for datasets as `EDDTableFromDapSequence` in the federating ERDDAP and recreating configs when the federated dataset changes just adds a ton of complexity/maintenance overhead and isn't feasible at our current scale. We'll find a path forward though. Thanks for the response!

BobSimons:
EDDTableFromDapSequence may be more complex than you need, but it still
seems like you could write a small program/script to make the changes you
need to make. (But I don't know the details of what you need to do.)

Another alternative is to make your own custom class which extends
EDDTableFromDapSequence or EDDTableFromErddap but just adds the options you
want. That would be more work up front, but would create a custom type of
dataset (and thus easy to use for your purposes).

Best wishes.

PR #22, opened 2018-12-13
Fixed OrderByMean NaN comparison to use Double.isNaN()

fullergalway:


BobSimons:
This was handled privately long ago.

PR #30, opened 2020-07-20
Discussion: speed up refresh for EDDTableFromMultidimNcFiles

jessicaaustin:
Hi @BobSimons - I'm opening a PR here mostly because it's the easiest way to discuss code changes -- but if you'd rather discuss this on the mailing list or as a github issue, I'm happy to move the conversation there instead.

**Background**

I'm trying to set up an ERDDAP server with datasets based on very large netcdf files. On the upper end, we're talking 50GB datasets that are composed of individual nc files that are 500MB each. 

I know that ERDDAP will take a while for the initial dataset load, since it's loading in the netcdf data and building its own table representation, and that's fine. However, after that initial load, I need to reload to be relatively fast, so that I don't have to wait hours or days for a reload if the server goes down and I need to restart. But this isn't what I was seeing. Here's an example:

* Initial load:
 `*** EDDTableFromFiles test-dataset_NC constructor finished. TIME=8701ms`
* Subsequent load (no modifications to the file at all in between; just restart ERDDAP):
 `*** EDDTableFromFiles test-dataset_NC constructor finished. TIME=4427ms`

This is only a 2x speed improvement -- and that's about what I was seeing for other test netcdf datasets, ranging in file size.

**Changes/Questions**

So I started debugging to see what was going on. There were three things I found that I wanted to ask you about. 

1. **`makeExpected`: update `lastMod` check**

    When reloading, `EDDTableFromFiles::makeExpected` checks to see if the file has changed since the last time it loaded. I found that there were cases where the file mod time was *less than* the last mod time that erddap had stored
    * example: `1588697736060 -- ftLastMod.get(f)` and `1588697736000 -- lastMod` -- only off by 60ms

    So i switched this check from `ftLastMod.get(f) == lastMod` to `ftLastMod.get(f) < lastMod`. For one of my test cases, this made it so that I went from around 5.5s for a reload of a 20MB nc file, to a .7s reload time, which makes sense -- if it thought the file had been modified, then it would do a full reload which is expensive.

    While I'm not sure why the file table and file last mod times are different at all, this seems like a reasonable fix? Thoughts?

2. **`EDDTableFromMultidimNcFiles`: `mustGetData` flag?**

   There is a flag, `mustGetData` that is passed to `lowGetSourceDataFromFile`. This flag is set to false during reload, because ERDDAP should not have to load up all the data at this point (after all, it's already checked that the file hasn't been modified since it last looked at it). However, that flag wasn't getting honored by `EDDTableFromMultidimNcFiles` (it isn't passed into `readMultidimNc`), so it was *loading up the data every single time*.  

    I started digging into this, but `readMultidimNc` is very complex so I didn't get very far. Plus my update (3) below made it so that `getSourceDataFromFile` is never called anyway. But I'm mostly curious why `EDDTableFromMultidimNcFiles` doesn't honor the `mustGetData` param -- it seems like updating it to do so would drastically improve reload speeds for netcdf file datasets.

3. **`makeExpected`: skip `getSourceDataFromFile`**
 
    After looking at the `mustGetData` issue, I was wondering why ERDDAP has to touch the nc file at all, if it hasn't been modified? In other words: in `EDDTableFromFiles.makeExpected`, if the file hasn't changed and it isn't the same size, then just completely skip loading up the `Table` for that file. 

    If I remove that code block (see changes) then it goes from 0.7s --> 0.2s for a reload. So overall, that's 5.5s-->0.2s, an over 90% decrease in time spent reloading. And I've run this branch with dozens of large datasets and saw a similar speedup, with no functionality problems that I could find.

    That said, I'm sure there is some reason that code block is in there? Are there other modifications that ERDDAP is looking for?

Thanks for reading, I look forward to hearing your responses.

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromFiles.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromFiles.java
@@ -1982,7 +1982,9 @@ public abstract class EDDTableFromFiles extends EDDTable{
             String name = ftFileList.get(f);
             if (filesAreLocal) {
                 long lastMod = File2.getLastModified(dir + name);
-                if (lastMod == 0 || ftLastMod.get(f) != lastMod) //0=trouble: unavailable or changed
+                // UPDATE:
+                // mod times don't have to be equal, just check if file has been modified since the table was last updated
+                if (lastMod == 0 || ftLastMod.get(f) < lastMod) //0=trouble: unavailable or changed
                     continue;
                 long size = File2.length(dir + name);
                 if (size < 0 || size == Long.MAX_VALUE || 
@@ -1990,29 +1992,9 @@ public abstract class EDDTableFromFiles extends EDDTable{
                     continue;
             }
 
-            try {
-                //get the metadata
-                //String2.log("here 1");
-                Table table = getSourceDataFromFile(dir, name,
-                    sourceDataNames, sourceDataTypes, 
-                    -1, Double.NaN, Double.NaN, 
-                    null, null, null, true, false); //getMetadata=true, getData=false
-                //String2.log("here 2");
-
-                //get the expected attributes;     ok if NaN or null
-                for (int dvNec = 0; dvNec < sourceDataNames.size(); dvNec++) {
-                    String tName = sourceDataNames.get(dvNec);
-                    int tableDv = table.findColumnNumber(tName);
-                    Attributes dvAtts = tableDv < 0? new Attributes() : table.columnAttributes(tableDv);
-                    expectedAddOffset[dvNec]    = dvAtts.getDouble("add_offset");  
-                    expectedFillValue[dvNec]    = dvAtts.getDouble("_FillValue");
-                    expectedMissingValue[dvNec] = dvAtts.getDouble("missing_value");
-                    expectedScaleFactor[dvNec]  = dvAtts.getDouble("scale_factor");
-                    expectedUnits[dvNec]        = dvAtts.getString("units");
-                }
-            } catch (Throwable t) {
-                throw new RuntimeException("Unexpected error when getting ExpectedXxx attributes from " + dir + name, t);
-            }
+            // UPDATE:
+            // If we got here, then the file hasn't been modified, and it's the same size.
+            // So don't even try to load it!
 
             //we got what we needed, no need to look at other files
             if (verbose) String2.log("ExpectedXxx attributes were read from " + dir + name);


BobSimons:
I understand the desire for a faster reload in quickRestart mode. That's the point of quickRestart mode.

To answer your questions:

1) The problem with switching to ftLastMod.get(f) < lastMod is: You're thinking "The file is 60ms older than expected. That's a quirk of the operating system, so let's ignore changes where the file is older than expected." But I'm thinking that an older file might be much older (e.g., 60 days older) because it is an older version of the file and therefore needs to be read and processed. So I am reluctant to make that change to the code.  

You could then say "well, then let's make ERRDAP ignore changes of <1 s. Perhaps. But It is possible that some real time dataset is changed very frequently, so it could be a different version of the file. So I am even reluctant to make that change.

I have no explanation of why the OS says the lastModified time is slightly different. That's disappointing. 

2)  The parent class EDDTableFromFiles deals with a wide range of subclasses, so the methods that the subclasses have to handle (notably lowGetSourceDataFromFile) sometimes include parameters that are requests, not obligations.  (It's actually more complicated than that, e.g., the whole CONSTRAIN_NO/PARTIAL/YES system.)  mustGetData (note that it isn't just called "getData") is thus stating a requirement that the implementation must get the data (otherwise it is optional). 

So why doesn't EDDTableFromMultidimNcFiles pay attention and use that parameter? You basically answered that: "readMultidimNc is very complex". I didn't take the time to add that option to the reader by adding to the complexity of an already complex method. 

3) The whole point of makeExpected is to get the expected values of certain vital metadata attributes from one of the source files that is known to be valid.  This information is used elsewhere to validate new or changed files to ensure that the variables have the same (expected) values of scale_factor, add_offset, missing_value, etc. so we know that aggregating all these files is a valid thing to do.  Most of the time, opening one file from a collection doesn't take long. But here it does. So we can't just remove that code.

There is a possible way around this: the dataset could store this information in a file in the dataset's info directory. But that adds considerable complexity to the process. So I am reluctant to do that.

Another option is: we could make makeExpected lazy. Even that might be somewhat complicated since I think the information is used in other places (e.g., updateEveryNMillis). I will investigate.

---
I feel bad. You did a lot of work to do the analysis and make the changes and I rejected your 3 suggested changes and left you without a solution. I'll say: the situation I envisioned for these datasets was: the aggregation consists of 1 to a million files, each of which can be read fairly quickly. When that is true, the time needed for makeExpected is minimal because it only ever opens 1 file, and thus quickRestart finishes quickly if no files have changed.  And I envisioned that restarting ERDDAP was a rare event. Your situation is different because reading one file takes significant time and you might have lots of datasets like this.

I'll mention an obvious solution: make the data files somewhat smaller. That has a few advantages. 
1) It solves this problem.
2) It makes it more likely that ERDDAP can read less data when responding to requests simply by quickly rejecting more of these smaller files. (But at the expense of sometimes having to open up lots more files. It's a trade-off.)
But I'm guessing you don't want to make that change. Fair enough.

I will try to think of other solutions. The ones I have thought of so far just add more complexity to an already fairly complex system. If you have other possible solutions, please let me know.



Another option is

jessicaaustin:
@BobSimons Thanks for the quick reply and thorough response!

> I feel bad. You did a lot of work to do the analysis and make the changes and I rejected your 3 suggested changes and left you without a solution.

You shouldn't feel bad; part of this was me validating I had set up my config properly. There are countless ways to configure an ERDDAP dataset and I wanted to understand what it was doing under the hood so I could make the right decisions about how to structure my datasets. With datasets this large, a small tweak can make all the difference. (Plus I was able to set up a unit test to try out different changes quickly, so the debugging was not as bad as you might think) 

I will say the only thing I found "surprising" was that even if my files did not change whatsoever, ERDDAP was still loading up all the data for the first file -- in other words that EDDTableFromMultidimNcFiles reads the data not just the metadata. 

> I'll say: the situation I envisioned for these datasets was: the aggregation consists of 1 to a million files, each of which can be read fairly quickly. When that is true, the time needed for makeExpected is minimal because it only ever opens 1 file, and thus quickRestart finishes quickly if no files have changed. 

Yes, this makes sense and is what I expected after looking at the code. So even if my dataset is 50GB, if it's broken up into very small files then the reload should be quick because it only has to look at one file to check the metadata against what's stored in the file table. I will think about this; there may be something I can do here, at least to make the first file smaller. 

> And I envisioned that restarting ERDDAP was a rare event. Your situation is different because reading one file takes significant time and you might have lots of datasets like this.

It should still be a rare event, but even if it's rare I need the reload to happen in a reasonable amount of time. For example, the [IOOS ESM ERDDAP](http://erddap.sensors.axds.co/erddap/index.html) takes about an hour to fully reload after restart, which means we do major config changes or upgrades on off hours, but they are still do-able.

> Another option is: we could make makeExpected lazy. Even that might be somewhat complicated since I think the information is used in other places (e.g., updateEveryNMillis). I will investigate.

I'm interested to hear where this goes. Would you consider adding a configuration flag that would make `makeExpected` more or less aggressive about how it does its updates? This could affect the `lastMod` check as well. So for example I would set that flag to "lazy" (if file not changed since last update, do nothing) for datasets with large files that I know change very infrequently, and "aggressive" (current behavior) for datasets that are updated very frequently and I want to make sure ERDDAP stays on top of those changes. 

> Another option is

Did you mean to have more here? It's cut off.

BobSimons:
Yes, the last "Another option is" was a typo.

I now suspect the best option is to make the code not get the data when mustGetData is false.  I'm super busy right now and the current release is super late, so I will probably put off working on this until the next release.

jessicaaustin:
Okay, that makes sense. I think as the size of datasets scale, this change will help more and more users, not just our particular use case here.

In the meantime, if I can't split into smaller files then I'll just run a fork with the `lastMod` change, and I can revert that once the netcdf mustGetData updates are in the main trunk a few releases from now.

Thanks!

shane-axiom:
> the best option is to make the [`EDDTableFromMultidimNcFiles` `lowGetSourceDataFromFile`] code not get the data when mustGetData is false

For reference, Bob implemented this change here, first released in v2.10:

https://github.com/BobSimons/erddap/commit/e55556f18d39d5ae3f4ff96748fec2ebd42ace63#diff-174e5ec59fb0f843a0e59a3ad65f2640ad372f2f226e2db95fb048d35dd49d35

PR #56, opened 2021-07-08
Add Chinese and German version of messages.xml, and a tool to produce the translation in other languages

Q1Zeng:
This is Qi Zeng's first PR for the GSoC project on translating ERDDAP's interface. It modifies the messages.xml slightly, and adds a rawContent feature to gov.noaa.pfel.coastwatch.util.SimpleXMLReader.xml. 
The translation tool and its dependencies are included. The example output, German and Chinese versions of messages.xml, are also uploaded.


Q1Zeng:
@ocefpaf @mwengren

PR #84, opened 2022-08-31
Fix capitalization inconsistency in TranslatedMessages xml files

mwengren:
The code in [TranslateMessages.java ](https://github.com/BobSimons/erddap/blob/master/WEB-INF/classes/gov/noaa/pfel/erddap/util/TranslateMessages.java#L211-L217) has different capitalization for one XML message file.  

This may not be an issue when running ERDDAP in Tomcat, but it presented a problem when testing in a different Servlet Container.


mwengren:
Closing, will be fixed in a future commit by @BobSimons

PR #97, opened 2023-03-28
Modified pom.xml to build ERDDAP WAR in place with mvn package

mwengren:
@ChrisPJohn These small pom.xml changes allow a version of the ERDDAP war file to be compiled via Maven in the `target` subdirectory by running:

```
mvn package
```

Or, step-by-step by running:

```
mvn resources:resources
mvn compile
mvn package
```

The resulting `target/ERDDAP-2.23-SNAPSHOT.war` file does not exactly match what is produced by [makeErddapWar.bat](https://github.com/BobSimons/erddap/blob/master/makeErddapWar.bat), but it's close enough for development purposes AFAICT.

Also, I wasn't sure how to handle the [netcdfAll-5.5.3.jar file issue in pom.xml](https://github.com/BobSimons/erddap/blob/master/pom.xml#L425-L427), but the simplest solution is to just reference it in place via relative path under `WEB-INF/lib` and not retain in pom.xml absolute paths that are specific to a local environment.  I can submit another PR with a different approach to resolve this that uses a Maven `settings.xml` file and also eliminates the absolute path in pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -98,17 +98,10 @@ since this is impossible for AWS S3 dependencies without Maven.
     </repositories>
 
     <build>
+        <sourceDirectory>WEB-INF/classes</sourceDirectory>
         <resources>
             <resource>
-                <directory>src/main/java</directory>
-                <includes>
-                    <include>**/*.properties</include>
-                    <include>**/*.csv</include>
-                    <include>**/*.tsv</include>
-                    <include>**/*.txt</include>
-                    <include>**/*.cpt</include>
-                    <include>**/*.xml</include>
-                </includes>
+                <directory>WEB-INF/classes</directory>
             </resource>
         </resources>
         <plugins>
@@ -145,9 +138,7 @@ since this is impossible for AWS S3 dependencies without Maven.
                         <exclude>gov/noaa/pfel/erddap/dataset/EDDTableFromNWISDV.javaINACTIVE</exclude>
                         <exclude>gov/noaa/pfel/erddap/dataset/EDDTableFromPostDatabase.javaINACTIVE</exclude>
                         <exclude>gov/noaa/pfel/erddap/dataset/EDDTableFromPostNcFiles.javaINACTIVE</exclude>
-                        
                         <exclude>gov/noaa/pfel/coastwatch/util/FileVisitorDNLS.javaOLD</exclude>
-
                     </excludes>
                 </configuration>
             </plugin>
@@ -157,6 +148,10 @@ since this is impossible for AWS S3 dependencies without Maven.
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-war-plugin</artifactId>
                 <version>3.3.2</version>
+                <configuration>
+                    <warSourceDirectory>${project.basedir}</warSourceDirectory>
+                    <warSourceExcludes>WEB-INF/lib/**, target/**, data/**, content/**, .settings/**, *.*</warSourceExcludes>
+                </configuration>
             </plugin>
 
             <!-- NOT ACTIVE. See above and below.
@@ -422,9 +417,7 @@ since this is impossible for AWS S3 dependencies without Maven.
             <artifactId>netcdfAll</artifactId>
             <version>5.5.3</version>  <!-- version number duplicated several times below -->
             <scope>system</scope>
-            <!-- Bob uses fixed path... -->
-            <systemPath>C:/programs/mavenERDDAP/ERDDAP/repo/edu/ucar/netcdfAll/5.5.3/netcdfAll-5.5.3.jar</systemPath> <!-- gets WARNING about absolute path, but okay -->
-            <!-- Chris uses <systemPath>${basedir}/netcdfAll-5.5.3.jar</systemPath> -->
+            <systemPath>${project.basedir}/WEB-INF/lib/netcdfAll-5.5.3.jar</systemPath>
         </dependency>
 
         <!-- These will be needed if I ever switch away from netcdfAll.
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ images/gadgets
 images/post
 *.class
 Thumbs.db
+/content/
+/data/
+/target/
 .vscode
 .settings
 .classpath


pom.xml
```diff
@@ -157,6 +148,10 @@ since this is impossible for AWS S3 dependencies without Maven.
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-war-plugin</artifactId>
                 <version>3.3.2</version>
+                <configuration>
+                    <warSourceDirectory>${project.basedir}</warSourceDirectory>
+                    <warSourceExcludes>WEB-INF/lib/**, target/**, data/**, content/**, .settings/**, *.*</warSourceExcludes>
```
ChrisPJohn, line 153: If I'm reading the makeErddapWar.bat properly, it includes WEB-INF/lib/ in the inputs. Is excluding this an intentional change?

mwengren: Yes, intentional, but it could be left in or taken out, it won't matter in the resulting WAR file generated.  Maven copies the same libraries in the repo to `target/${project.build.finalName}/WEB-INF/lib` based on the dependency configuration in pom.xml if `WEB-INF/lib` is in the exclude list here.  I've tested several times to confirm.  

Actually, the `WEB-INF/lib` directory in the ERDDAP repo could have all JAR files removed, except `netcdfAll-5.5.3.jar` (since it's not available elsewhere any longer) if desired.  Then pom.xml would be entirely responsible for determining dependencies and it would eliminate the duplication.  

I think this is better myself, but it's not essential, this PR can still be merged without any changes made to `WEB-INF/lib` if there's reason to keep those JARs in the source code.

BobSimons:
The ERDDAP repository used to include everything needed to install and run ERDDAP. Is it a good idea to move away from that?  

We are introducing a dependency on installing and running Maven for everyone who wants to use the ERDDAP repo. Is that a good idea? I know Maven seems like a standard tool these days, but I've seen a lot of tools come and go over the years. People used to chide me for not using Subversion. Many projects still require Eclipse. I didn't want to force developers to use any specific 3rd party tools, especially because there were times when different projects required different competing tools (e.g., Maven and whatever was before Maven for builds). The only reason I was, or Chris is, forced to use Maven was/is to get all the AWS dependencies. But other developers don't have to do that. Isn't it better to just have the needed libraries in the repo? Chris, I leave it to you, but I am not keen on requiring other developers to use Maven when it isn't necessary for them to work on ERDDAP. It's an extra step (a big black box) and one more thing to go wrong.  Maven has been 95% trouble for me. (Chris, one of the To Do list items is use Maven to get netcdf included in ERDDAP via its components instead of netcdf-All. I worked on that a lot and failed. The netcdf-java people were unable to help.) Maven is great when it does what you want and terrible when it doesn't.


And Micah, you say "The resulting target/ERDDAP-2.23-SNAPSHOT.war file does not exactly match what is produced by makeErddapWar.bat), but it's close enough for development purposes AFAICT."  What are the differences and what is your basis for saying that is close enough?

BobSimons:
Stated another way: Chris, if you accept this change that makes Maven a requirement for using the ERDDAP repo, you are effectively signing up to help anyone who has problems running Maven and getting it to produce the desired results. Is that a good idea?   (Maybe it is okay because you are confident it won't be a problem.) And what is the huge (or tiny) benefit that justifies taking on this responsibility? I see none (other than maybe allowing that one big file to be automatically downloaded and installed in the repo, allowing the repo to be 30% smaller; there are other ways to accomplish that). What am I missing?

mwengren:
@BobSimons Some responses to your comments.

> We are introducing a dependency on installing and running Maven for everyone who wants to use the ERDDAP repo. Is that a good idea? 

This PR doesn't actually change ERDDAP's dependency on Maven at all, it only fixes the existing pom.xml in a few places to enable building ERDDAP with Maven if desired.  Since the pom.xml is already part of this project, there should be no harm in making it more correct for ERDDAP's source structure.  

All of the existing libraries can be kept in `WEB-INF/lib` - this PR doesn't make any changes there, I only mention it for future consideration in response to @ChrisPJohn's comments.

> And Micah, you say "The resulting target/ERDDAP-2.23-SNAPSHOT.war file does not exactly match what is produced by makeErddapWar.bat), but it's close enough for development purposes AFAICT." What are the differences and what is your basis for saying that is close enough?

The reason I say it's close enough is that I'm able to run ERDDAP in a 'development mode' that's easier to configure and faster for iterative development than relying on a full-blown [Tomcat installation](https://coastwatch.pfeg.noaa.gov/erddap/download/setup.html#initialSetup).  

Unfortunately, I can't run the test framework to know definitively if there's a functional difference between the two, but the differences in archive content are minimal - mostly the resource files that are/are not included from [here](https://github.com/BobSimons/erddap/blob/master/makeErddapWar.bat#L14).  

Building with Maven using the changes in this PR basically causes all of the non-Java files in `WEB-INF` to be packaged in the war file.  I can tell that's not ideal, but I'm not suggesting Maven replace the production ERDDAP build process, to be clear.  

That said, it should be possible to replicate the file list from makeErddapWar.bat in pom.xml if that were desired, but I haven't taken that step so far because ERDDAP has an existing build process.

> Chris, if you accept this change that makes Maven a requirement for using the ERDDAP repo, you are effectively signing up to help anyone who has problems running Maven and getting it to produce the desired results.

> And what is the huge (or tiny) benefit that justifies taking on this responsibility?

I don't think the first statement is true, at least for this PR anyway.    It will/would be in relation to my other PR #98, so I think it would be better to have that discussion there - or I can start a discussion topic instead and the second PR can be on hold - mostly I wanted to share that code with Chris per yesterday's conversation.

My answer to the last question is that I think enabling ERDDAP to be run in a Maven-based development mode without a Tomcat dependency is actually a huge benefit that should make any prospective ERDDAP developer more productive, and worth the tradeoff off of more fully adopting Maven.

I think the best way to demonstrate that though is to just submit another PR that will allow anyone that wants to to try it for themselves (or I can update #98 instead).   I know from my own Java experience that developing web applications with Tomcat slower than I wanted it to be, so I looked for other options and was happy that I did.

ChrisPJohn:
>The ERDDAP repository used to include everything needed to install and run ERDDAP. Is it a good idea to move away from that?

As Micah mentioned this change doesn't effect that currently. That said the common practice with dependency management tools available for software development is to have a configuration (like pom.xml) that defines your dependencies and a tool (like maven) that fetches and installs those dependencies. This allows for a consistent environment setup for developers while reducing the size of artifacts needing to be stored in source control.

>We are introducing a dependency on installing and running Maven for everyone who wants to use the ERDDAP repo. Is that a good idea? I know Maven seems like a standard tool these days, but I've seen a lot of tools come and go over the years. People used to chide me for not using Subversion. Many projects still require Eclipse. I didn't want to force developers to use any specific 3rd party tools, especially because there were times when different projects required different competing tools (e.g., Maven and whatever was before Maven for builds). The only reason I was, or Chris is, forced to use Maven was/is to get all the AWS dependencies. But other developers don't have to do that. Isn't it better to just have the needed libraries in the repo? Chris, I leave it to you, but I am not keen on requiring other developers to use Maven when it isn't necessary for them to work on ERDDAP. It's an extra step (a big black box) and one more thing to go wrong. Maven has been 95% trouble for me. (Chris, one of the To Do list items is use Maven to get netcdf included in ERDDAP via its components instead of netcdf-All. I worked on that a lot and failed. The netcdf-java people were unable to help.) Maven is great when it does what you want and terrible when it doesn't.

> Stated another way: Chris, if you accept this change that makes Maven a requirement for using the ERDDAP repo, you are effectively signing up to help anyone who has problems running Maven and getting it to produce the desired results. Is that a good idea? (Maybe it is okay because you are confident it won't be a problem.) And what is the huge (or tiny) benefit that justifies taking on this responsibility? I see none (other than maybe allowing that one big file to be automatically downloaded and installed in the repo, allowing the repo to be 30% smaller; there are other ways to accomplish that). What am I missing?

Maven is not a new tool, it's been around for over 20 years and is widely supported and used in software development. I'm not concerned with an increased support load in ERDDAP by using it. In fact, I expect by using a industry standard tool it will make it easier for developers new to ERDDAP to understand what is going on. This could eventually include running tests through maven to make that easier for new developers and automating many of the currently manual processes.  Also, I agree with Micah, that being able to run ERDDAP without Tomcat locally for development will be a huge benefit (of course I'm also very aware that before any release changes should be checked through Tomcat as well). This all makes ERDDAP development more accessible to new developers and reduces barriers in my opinion.

BobSimons:
Sorry. Yes, my comments are about the other pull request. I responded to
them here (and continue to respond to them here) because you discussed them
here.

You say it's "easier to configure and faster for iterative development than
relying on a full-blown Tomcat installation
<https://coastwatch.pfeg.noaa.gov/erddap/download/setup.html#initialSetup>".
I guess that is true if you prefer "compile" in Maven to "javac" in a
programmer's editor (is there some advantage I don't know about?) or if you
have tests set up in Maven (we don't), otherwise I don't understand. A
Tomcat installation is useful because it is an easy way to set things up
the standard way for ERDDAP (e.g., setup.xml and datasets.xml) and to test
anything related to web services (which is the main point of ERDDAP). If
you have a Tomcat installation and are working on something that is not a
web service (e.g., a file reader), it is easy to do quick iterations with
javac and java (i.e., not restarting or even using Tomcat).as you would
with any other Java application. (You seem to disagree with that. Why?)

Stated another way, the natural home for ERDDAP is in Tomcat, and is not
(nor will it ever be) in Maven.  As I said before, the only person that
needs to use Maven for ERDDAP was me and is now Chris, so that we can
update the /lib libraries. We'll always have to support ERDDAP in Tomcat,
so why add another option that adds little (or nothing) and is just another
thing to go wrong?

I see you love Maven and it appears to be your favorite work environment.
Fair enough. But from an ERDDAP support standpoint, that's not a great
reason to favor Maven as the development environment as it isn't useful
when working with Tomcat. There are lots of Java developers who love
Eclipse (with good reason). Does that mean we should gear deployment to
Eclipse if someone requests it? I say no. I think it is better for ERDDAP
to be as tool-agnostic as possible. If you want to use Maven for
development, you can. If someone else wants to use Eclipse, they can.

But again, I leave the decision to Chris.

ChrisPJohn:
I'm going to accept this pull request since it enables a new workflow for those that want to use it, but doesn't impact any existing workflows. Any additional discussion about the changes in #98 should be moved to that pull request.

PR #101, opened 2023-04-15
Remove large files from the repo, add support to maven to download them

ChrisJohnNOAA:
The general approach is that maven will manage code dependencies (including copying them to the WEB-INF/lib directory to minimize differences from existing workflows). Maven is also used to download other needed files like those in the WEB-INF/ref directory.

This means these large files no longer need to be in the ERDDAP repo, but with maven there is still file versioning available.

This does add a step for developers to load the dependencies using maven, but that is a pretty standard step for most large software projects.


pom.xml
```diff
@@ -105,6 +70,32 @@ since this is impossible for AWS S3 dependencies without Maven.
...
+            <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>install</phase>
+                        <goals>
+                            <goal>copy-dependencies</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>WEB-INF\lib</outputDirectory>
```
mwengren: Suggest keeping the path separator syntax consistent `WEB-INF/lib`. Here and on L93.

pom.xml
```diff
@@ -154,39 +145,86 @@ since this is impossible for AWS S3 dependencies without Maven.
...
+                        <?m2e execute onConfiguration?>
+                        <id>download-erddap-content</id>
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <get src="https://github.com/BobSimons/erddap/releases/download/${erddapcontent.download.version}/erddapContent.zip" dest="${project.basedir}/erddapContent.zip" skipexisting="true" verbose="true"/>
+                                <!-- note: the overwrite="false" option doesn't seem to act as advertised.  It will overwrite files in the dest folder if they are older than in archive -->
```
mwengren: When I last ran this, I'm pretty sure it did not overwrite my local setup.xml file (even though it had not been edited more recently than the most [recent commit](https://github.com/BobSimons/erddapContent/commit/711ecd9791ae5543e909c29351d71f900bf4eda6) on erddapContent of setup.xml.  So, I'd say remove this warning here and on the lines below unless you or others happen to notice the overwrite behavior.  In practice, it really probably only matters for setup.xml and other customized files in content/erddap directory (such as datasets.xml) since the other ref files below are aren't likely to be modified.

pom.xml
```diff
@@ -154,39 +145,86 @@ since this is impossible for AWS S3 dependencies without Maven.
...
+                    <execution>
+                        <?m2e execute onConfiguration?>
+                        <id>download-erddap-content</id>
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <get src="https://github.com/BobSimons/erddap/releases/download/${erddapcontent.download.version}/erddapContent.zip" dest="${project.basedir}/erddapContent.zip" skipexisting="true" verbose="true"/>
```
mwengren: I would save this locally under `download_cache` as you have for the other files for consistency, and skip the delete step on L179 as well.

ChrisJohnNOAA:
> 

Thanks for the comments. I've addressed those changes and made some fixes that were needed to make the process work smoothly when starting from a new clone of the repo as well as updated the setup.html documentation.

pom.xml
```diff
@@ -416,8 +467,6 @@ since this is impossible for AWS S3 dependencies without Maven.
             <groupId>edu.ucar</groupId>
             <artifactId>netcdfAll</artifactId>
             <version>5.5.3</version>  <!-- version number duplicated several times below -->
-            <scope>system</scope>
-            <systemPath>${project.basedir}/WEB-INF/lib/netcdfAll-5.5.3.jar</systemPath>
```
mwengren, line 420: I haven't tested with the system scope and systemPath configs removed like this, but if it works for you, then :+1: for dropping them here.  

This results in the netcdfAll jar being packaged with the WAR file when built with the standard Maven build, right?  Is that desired?

mwengren (approved):
Looks good to me, save a couple comments below.

pom.xml
ChrisJohnNOAA: Yeah, in my testing the netcdf jar is included in the war without these lines. Worst case if it doesn't work reliably, we can re-add them.

ChrisPJohn (approved):

PR #116, opened 2023-11-03
Alter pom.xml to stage war in target/war-staging for test execution

srstsavage:
Also demonstrate altering setup.xml values to set up suitable test values, add simple example tests, and separate execution of tests with remote dependencies to a separate maven profile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ images/post
 /data/
 /target/
 download_cache/
+tmp/
 *.class
 Thumbs.db
 .vscode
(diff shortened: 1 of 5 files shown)


srstsavage:
This PR was created mainly to inform discussion on how Axiom handles adding junit tests to the existing upstream ERDDAP codebase with minimal modifications: stage the build assets in a directory structure simulating the Tomcat environment so that assumptions about relative locations etc are met, and altering a setup.xml file enough to allow ERDDAP (`EDStatic`) to initialize.

srstsavage:
Closing this PR as it was meant to demonstrate one possible approach for minimal refactoring to accommodate test classes. jUnit tests were added starting mostly in https://github.com/ERDDAP/erddap/pull/142.

Approaches in this PR can be revisited as needed, but no need to keep this open IMO.

PR #121, opened 2023-11-29
Move one test into JUnit as an example

ChrisJohnNOAA:


pom.xml
```diff
@@ -234,6 +235,10 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+				<artifactId>maven-surefire-plugin</artifactId>
+				<version>3.2.1</version>
+			</plugin>
```
srstsavage: Whitespace/indenting is a bit wonky here.

pom.xml
```diff
@@ -287,8 +280,15 @@
...
             </dependency>
-        </dependencies>
 
+            <dependency>
+				<groupId>org.junit</groupId>
+				<artifactId>junit-bom</artifactId>
+				<version>5.10.1</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
```
srstsavage: Again, indenting needs fixin

pom.xml
```diff
@@ -306,16 +306,11 @@
...
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.13.2</version>
-            <scope>test</scope>
-        </dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter</artifactId>
+			<scope>test</scope>
+		</dependency>
```
srstsavage: Indenting :)

src/test/WEB-INF/classes/com/cohort/array/ByteArrayTests.java
```diff
@@ -0,0 +1,428 @@
...
+        anArray.setLong(0, 4);              Test.ensureEqual(anArray.getLong(0),   4, ""); 
+        anArray.setInt(0,Integer.MAX_VALUE);Test.ensureEqual(anArray.getInt(0),    Integer.MAX_VALUE, ""); 
+        anArray.setInt(0, 1123456789);      Test.ensureEqual(anArray.getInt(0),    Integer.MAX_VALUE, ""); 
+        anArray.setInt(0, 5);               Test.ensureEqual(anArray.getInt(0),    5, ""); 
+
+        //makeUnsignedPA
+        anArray = new ByteArray(new byte[] {-128, -2, -1, 0, 1, 126, 127});
+        UByteArray uArray = (UByteArray)anArray.makeUnsignedPA();
+        Test.ensureEqual(uArray.toString(), "128, 254, 255, 0, 1, 126, 127", ""); // -1 -> mv
+        anArray.clear();
```
srstsavage: This line and many others in this test have trailing whitespace. Admittedly this is a nit of infinitesimal importance, but if we're moving code around it might be a nice opportunity to clean that up.

srstsavage (changes requested):
Awesome! I can verify that the example test passes using the expected/standard `mvn clean test`.

Added a couple of nits about whitespace/indenting, but :+1: to the overall approach

srstsavage:
Looks good! Can you comment on the approach of using `src/test/WEB-INF/classes` as the test dir instead of the standard `src/test/java`?

srstsavage (approved):
Looks good!

Regarding the path for the tests, my personal recommendation would be to just use `src/test/java` (so `src/test/java/com/cohort/array/ByteArrayTests.java` for this example) because that's the normal convention (package dir hierarchy under `src/test/java` and because the web application context/normal usage is unrelated to the testing of the individual classes. But, the current `src/test/java/WEB-INF` directory works as well, and there's more than one way to lay out a project etc. No protests from me if you'd prefer to keep it as-is.

srstsavage (approved):

PR #123, opened 2023-12-27
Update setup.html

noaaroland:
Add PMEL to the list if institutions using ERDDAP with a link to our main server.

### DIFF
--- a/download/setup.html
+++ b/download/setup.html
@@ -283,6 +283,9 @@ Joint Research Centre of the European Commission)
       title="This link to an external website does not constitute an endorsement."></a> 
   (Pacific Islands Fisheries Science Center) 
   <!-- 2019-07-15 Melanie Abecassis -->
+<li><a rel="bookmark" href="https://data.pmel.noaa.gov/pmel/erddap/index.html">NOAA PMEL<img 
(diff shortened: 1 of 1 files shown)


ChrisJohnNOAA:
Would you be able to make this edit in the repo for the ERDDAP github.io pages:
https://github.com/ERDDAP/erddap.github.io/blob/main/setup.html

That repo is what generates the pages at: https://erddap.github.io/

Thanks!

noaaroland:
Sure. I did so. Does this still go out with each distribution so that if somebody wanders across a version [erddap siite]/erddap/download/setup.html (e.g. https://data.pmel.noaa.gov/pmel/erddap/download/setup.html) then this is source of that page?

ChrisJohnNOAA:
> Sure. I did so. Does this still go out with each distribution so that if somebody wanders across a version [erddap siite]/erddap/download/setup.html (e.g. https://data.pmel.noaa.gov/pmel/erddap/download/setup.html) then this is source of that page?

My intention is to eventually have general ERDDAP documentation in the github pages (or a similar centrally located documentation site). Until a plan is in place for that transition, it's best to keep both up to date.

ChrisJohnNOAA (approved):

PR #126, opened 2024-01-16
Create feature_request.yml

ChrisJohnNOAA:
Adding a template to guide feature requests to discussions

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: New Feature Request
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: New Feature Request
+    url: https://github.com/ERDDAP/ERDDAP/discussions/new
```
abkfenris, line 4: Instead of sending users right to creating a new discussion, you might want to link to the discussion category, so they at least are given the nudge to look through and see if anyone else is already discussing the same idea, before they create a new thread.

PR #141, opened 2024-03-29
Fix issue #119: Incorrect varName in logs

### LINKED ISSUE #119: Inaccurate varName in logs when printing for null value attribute
When looking in the log.txt file for problems at loading time, it seems there is a small issue when printing the variable attributes that have a null value. Here is an example of what I got:
```
Warning: varName=units attribute=units has values=null
Warning: varName=sdn_uom_urn attribute=sdn_uom_urn has values=null
Warning: varName=sdn_uom_name attribute=sdn_uom_name has values=null
Warning: varName=sdn_uom_name attribute=sdn_uom_name has values=null
Warning: varName=units attribute=units has values=null
```

As you can see, the "varName" actually takes the same value as the "attribute". 
I have erddap version 2.23.

ayushsingh01042003:
# Description

`varName` was being incorrectly logged by being copied from the `attribute`

Fixes #119 

## Type of change

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/coastwatch/griddata/NcHelper.java
+++ b/WEB-INF/classes/gov/noaa/pfel/coastwatch/griddata/NcHelper.java
@@ -1343,11 +1343,12 @@ public class NcHelper  {
         //read the sourcglobalAttributes
         if (ncAtts == null)
             return;
+        String ncAttsName = ncAtts.getName();
         Iterator it = ncAtts.iterator();
         while (it.hasNext()) { //there is also a dods.dap.Attribute                        
             Attribute att = (Attribute)it.next();            
-            String name = ncAtts.getName();
-            attributes.add(name, getAttributePA(name, att));
+            String name = att.getName();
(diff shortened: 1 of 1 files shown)


ayushsingh01042003:
Hi @ChrisJohnNOAA I've raised the PR as instructed, please take a look if this works and let me know if otherwise.

ayushsingh01042003:
@ChrisJohnNOAA That was careless on my part I completely overlooked the name parameter for attributes.add().

Just to make sure I understand the requested changes, The solution should look something like this?
```
  String ncAttsName = ncAtts.getName();
  Iterator it = ncAtts.iterator();
  while (it.hasNext()) { //there is also a dods.dap.Attribute                        
      Attribute att = (Attribute)it.next();            
      String name = att.getName();
      attributes.add(name, getAttributePA(ncAttsName, att));
  }
```
I think this should do it, I have pushed the changes, Can you have look at it?

ChrisJohnNOAA (approved):

PR #181, opened 2024-07-30
Sax Parser: Additional Handlers and Tests

ayushsingh01042003:
### Description
- AddAttributes Handler, AxisVariable Handler, DataVariable Handler
- EDDGridFromDap, EDDTableFromEDDGrid, EDDTableFromErddap, EDDGridFromErddap, EDDGridLonPM180 Handlers.
- Tests for the above


WEB-INF/classes/gov/noaa/pfel/erddap/handlers/DataVariableHandler.java
```diff
@@ -0,0 +1,52 @@
...
+import com.cohort.util.String2;
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+
+import java.util.ArrayList;
+
+public class DataVariableHandler extends State {
+    StringBuilder content = new StringBuilder();
+    String tSourceName = null, tDestinationName = null, tDataType = null;
+    com.cohort.array.Attributes tAttributes = null;
```
ChrisJohnNOAA: This looks like you never assign to it.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/AxisVariableHandler.java
```diff
@@ -0,0 +1,86 @@
...
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+
+import java.util.ArrayList;
+
+public class AxisVariableHandler extends State {
+    private StringBuilder content = new StringBuilder();
+    private ArrayList tAxisVariables;
+    private String tSourceName = null, tDestinationName = null;
+    private com.cohort.array.Attributes tAttributes = null;
```
ChrisJohnNOAA: This looks like you never assign to it.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/AddAttributesHandler.java
```diff
@@ -0,0 +1,79 @@
...
+
+public class AddAttributesHandler extends State {
+    private StringBuilder content = new StringBuilder();
+    private com.cohort.array.Attributes tAttributes = new com.cohort.array.Attributes();
+    private String tName = null, tType =  null;
+    private State completeState;
+
+    public AddAttributesHandler(SaxHandler saxHandler, com.cohort.array.Attributes tAddGlobalAttributes, State completeState) {
+        super(saxHandler);
+        tAddGlobalAttributes = this.tAttributes;
```
ChrisJohnNOAA: Is this supposed to be `this.tAttributes = tAddGlobalAttributes;`

As this is written I believe this line has no effect.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridLonPM180Handler.java
```diff
@@ -0,0 +1,122 @@
...
+                                    + "The dataset defined in an "
+                                    + "EDDGridLonPM180 must be a subclass of EDDGrid.");
+                }
+
+                String active = attributes.getValue("active");
+                String childDatasetID = attributes.getValue("datasetID");
+                State state = HandlerFactory.getHandlerFor(tType, childDatasetID, active, this, saxHandler, context);
+                saxHandler.setState(state);
+            } else {
+                throw new RuntimeException(
```
ChrisJohnNOAA: I think this else is actually if there's a second child dataset. Should the message be different?

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridLonPM180Handler.java
```diff
@@ -0,0 +1,122 @@
...
+                        "Datasets.xml error: "
+                                + "The dataset defined in an "
+                                + "EDDGridLonPM180 must be a subclass of EDDGrid.");
+            }
+        } else {
+            String2.log("Unexpected start tag: " + localName);
```
ChrisJohnNOAA: If no other start element is expected, why does the endElement have so many tags it supports?

ChrisJohnNOAA (changes requested):
Exciting to see parsers for so many dataset types. I had a few comments on the changes.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/DataVariableHandler.java
ayushsingh01042003: If we take a look at line 25 in the same file. I am passing tAttributes as an argument to the constructor
`new AddAttributesHandler(saxHandler, tAttributes, this);`
This then adds the parsing results from `AddAttributesHandler` to tAttributes

This is how the constructor looks
`public AddAttributesHandler(SaxHandler saxHandler, com.cohort.array.Attributes tAddGlobalAttributes, State completeState)`

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/AxisVariableHandler.java
ayushsingh01042003: Again similar to the above its passed as an argument to the constructor on line 29,
`State state = new AddAttributesHandler(saxHandler, tAttributes, this);`

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/AddAttributesHandler.java
ayushsingh01042003: So `tAddGlobalAttributes` is actually null and `this.tAttributes` is pointing to the object which has the parsing results needed which is initialised a few lines above this.

As for why is it named `tAddGlobalAttributes` and not just `tAttributes`.
If we take a look at a class like `EDDGridFromDap` you can see a line like,
```
else if (localTags.equals("<addAttributes>"))
        tGlobalAttributes = getAttributesFromXml(xmlReader);
```

But maybe changing the argument name to tAttributes might be better here?
That would make the line look like `tAttributes = this.tAttributes`

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridLonPM180Handler.java
ayushsingh01042003: Yes that's right, there is a different message for this condition. I missed it, will change it.
Thanks for pointing it out!

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridLonPM180Handler.java
ayushsingh01042003: Ah. I think we might have a big problem now that I notice.
Since the function was String2.log() I did not notice it till now but on using println() the unexpected tag message shows up. 

And this is a pattern I have used across all the handlers which means that I would either have to go back and add a bunch of empty conditions for the `startElement` function.

Or another option might be to simple get rid of the `String2.log("Unexpected start tag: " + localName);` line from the `default` and `else` conditions.
I believe this is better. Since if there is an unnecessary tag in there it will simple be ignored.
What do you think?

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/AddAttributesHandler.java
ayushsingh01042003: So `tAddGlobalAttributes` is passed via handlers like `EDDGridFromDapHandler`, It is initially null and we pass it to the AddAttributes handler and make it point to `this.tAttributes` which is initialised above and has the final object we need after parsing.

I however think maybe just name it `tAttributes` instead of `tAddGlobalAttributes` in the AddAttributesHandler class. Should I change it?

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/DataVariableHandler.java
ChrisJohnNOAA: Java passes by reference value, so re-assigning a parameter inside the called method will have no effect outside of the method. (Modifying the object does have an effect, but assignment is different, it's not modification). This stack overflow comment has a little more explanation: https://stackoverflow.com/questions/2229389/is-it-problematic-to-assign-a-new-value-to-a-method-parameter/2229479#2229479

This applies here and to the comments on AxisVariableHandler, and AddAttributesHandler.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridLonPM180Handler.java
ChrisJohnNOAA: It seems that endElement has code supported all (or almost all) of the elements a dataset expects. It might make more sense for the logging of unexpected elements to only be in the endElement function.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/DataVariableHandler.java
ayushsingh01042003: Oh thats right because I assigned `tAttributes` to null, its not the reference but null being passed in the constructor. I hope I got that correct?

So simply not assign it to null then?

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridLonPM180Handler.java
ayushsingh01042003: Yeah I think that's true. Will make that change.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/DataVariableHandler.java
ChrisJohnNOAA: No.

Java passes parameters by value. When you provide an object as a parameter, the value that is passed is the value of the reference for the object. What this means is changes to the internals of the object can be seen by the caller. However re-assigning the object has no impact in the caller.

Here is a longer explanation: https://www.digitalocean.com/community/tutorials/java-is-pass-by-value-and-not-pass-by-reference

ayushsingh01042003:
@ChrisJohnNOAA . Pushed some changes for the attributes object. Could you take a look and let me know if this is what you were reffering to?

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/AxisVariableHandler.java
```diff
@@ -0,0 +1,85 @@
...
+import com.cohort.util.String2;
+import java.util.ArrayList;
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+
+public class AxisVariableHandler extends State {
+  private StringBuilder content = new StringBuilder();
+  private ArrayList tAxisVariables;
+  private String tSourceName = null, tDestinationName = null;
+  private com.cohort.array.Attributes tAttributes;
```
ChrisJohnNOAA: I believe this needs to get initialized.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridLonPM180Handler.java
```diff
@@ -0,0 +1,125 @@
...
+            HandlerFactory.getHandlerFor(tType, childDatasetID, active, this, saxHandler, context);
+        saxHandler.setState(state);
+      } else {
+        throw new RuntimeException(
+            "Datasets.xml error: "
+                + "There can be only one <dataset> defined within an "
+                + "EDDGridLonPM180 <dataset>.");
+      }
+    } else {
+      String2.log("Unexpected start tag: " + localName);
```
ChrisJohnNOAA: You still need to remove these "Unexpected start tag" logs here and elsewhere.

ChrisJohnNOAA (changes requested):
That's better fort handling the attribute parameter, but I think you missed a spot to instantiate the attributes.

ayushsingh01042003:
Ran the Tests, no issues there. How does this look?

ChrisJohnNOAA (approved):
Awesome, thanks for the fixes!

mwengren:
I've been following this by email a little.  Looks great!  Please add the GSoC label (see email for details) to this and other GSoC-related PRs so we can track @ayushsingh01042003 's contributions!

PR #186, opened 2024-08-06
Handlers

ayushsingh01042003:
### Description

- EDDTableFromFilesHandler, EDDTableFromAsciiFilesHandler
- EDDGridAggregateExistingDimensionHandler
- EDDGridCopyHandler
- EDDGridFromEDDTableHandler
- EDDGridLon0360Handler

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridCopyHandler.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridCopyHandler.java
@@ -3,6 +3,7 @@ package gov.noaa.pfel.erddap.handlers;
 import static gov.noaa.pfel.erddap.dataset.EDDGrid.DEFAULT_MATCH_AXIS_N_DIGITS;
 
 import com.cohort.array.StringArray;
+import com.cohort.util.SimpleException;
 import com.cohort.util.String2;
 import gov.noaa.pfel.erddap.dataset.EDD;
 import gov.noaa.pfel.erddap.dataset.EDDGrid;
@@ -47,6 +48,9 @@ public class EDDGridCopyHandler extends State {
   public void startElement(String uri, String localName, String qName, Attributes attributes)
       throws SAXException {
     if (localName.equals("dataset")) {
+      if (tSourceEdd != null) {
+        throw new SimpleException("Cannot Have multiple Child datasets for" + datasetID);
+      }
       String tType = attributes.getValue("type");
       String active = attributes.getValue("active");
       String childDatasetID = attributes.getValue("datasetID");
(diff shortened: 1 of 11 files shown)


WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridCopyHandler.java
```diff
@@ -0,0 +1,121 @@
...
+  private String tDefaultGraphQuery = null;
+  private int tnThreads = -1;
+  private boolean tAccessibleViaFiles = EDStatic.defaultAccessibleViaFiles;
+  private boolean tDimensionValuesInMemory = true;
+  private String tOnlySince = null;
+
+  @Override
+  public void startElement(String uri, String localName, String qName, Attributes attributes)
+      throws SAXException {
+    if (localName.equals("dataset")) {
```
ChrisJohnNOAA, line 50: Should this throw an exception if it sees a second dataset?

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridFromEDDTableHandler.java
```diff
@@ -0,0 +1,162 @@
...
+  private int tnThreads = -1; // interpret invalid values (like -1) as EDStatic.nGridThreads
+  private boolean tDimensionValuesInMemory = true;
+  private EDDTable tEDDTable = null;
+  private int tGapThreshold = 1000;
+
+  @Override
+  public void startElement(String uri, String localName, String qName, Attributes attributes)
+      throws SAXException {
+    switch (localName) {
+      case "dataset" -> {
```
ChrisJohnNOAA, line 55: Should this throw an exception if it sees a second dataset?

ChrisJohnNOAA (changes requested):
Maybe some additional handling of incorrect xml noted in the comments.

WEB-INF/classes/gov/noaa/pfel/erddap/handlers/EDDGridCopyHandler.java
ayushsingh01042003: I don't think the old code specifies throwing an exception on getting a second child dataset,
https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGridFromEDDTable.java#L113-L127

But I am guessing since the `tEDDTable` variable will update to the last child dataset which means there is no point in adding multiple, you are implying to add a check so that we only have one child dataset?

ChrisJohnNOAA: Yeah, add a check to make sure there's only one child dataset. Allowing multiple to be parsed could cause issues, and I think it's better to catch those early with explicit error messages.

ayushsingh01042003:
How does this look?
I think there are handlers that would need this (all unless there are some that can have multiple child datasets)
If this looks fine I can make the changes to the other ones too.

ChrisJohnNOAA (approved):

srstsavage:
Don't forget the `GSoC-2024` label :grin:

PR #188, opened 2024-08-07
Add test datasets using Zarr files

ChrisJohnNOAA:
# Description

This adds test datasets based on Zarr files and tests to verify behavior.

Because zarr "files" are different (they're folders), there is some subtle behavior differences needed to make them work in ERDDAP. This approach requires datasets to include "zarr" in the filename or path regex for the dataset to trigger the behavior differences. It might be simpler for users (though likely much more code and/or more complicated changes) to have a new dataset type(s) that are specific to zarr files.

## Type of change

- [ X ] New feature (non-breaking change which adds functionality)

- [ X ] I have performed a self-review of my code
- [ X ] My code follows the style guidelines of this project
- [ X ] I have commented my code, particularly in hard-to-understand areas
- [ X ] I have added tests that prove my fix is effective or that my feature works
- [ X ] New and existing unit tests pass locally with my changes


ocefpaf:
> Because zarr "files" are different (they're folders)

Some people store them as zipped zarr files (the same folder but zipped). Not sure if we want to support that in ERDDAP.

ChrisJohnNOAA:
> > Because zarr "files" are different (they're folders)
> 
> Some people store them as zipped zarr files (the same folder but zipped). Not sure if we want to support that in ERDDAP.

I'm looking into this. There are additional changes needed to support zipped zarr files.

ChrisJohnNOAA:
> > > Because zarr "files" are different (they're folders)
> > 
> > 
> > Some people store them as zipped zarr files (the same folder but zipped). Not sure if we want to support that in ERDDAP.
> 
> I'm looking into this. There are additional changes needed to support zipped zarr files.

I added zipped zarr support.

PR #189, opened 2024-08-10
Wrapping up all Sax Parser Handlers

ayushsingh01042003:
### Description
- This completes all the available handlers(except Etopo)


ChrisJohnNOAA (approved):

ChrisJohnNOAA:
> ### Description
> * This completes all the available handlers(except Etopo)

We do need the etopo datasets. Is there a complication with doing them, or are they just not in this pull?

ayushsingh01042003:
Was not sure initially how to handle a self closing tag but turns out SAX treats it as both - opening and closing tag.
Pushed Etopo too.
This probably wraps up all the handlers.

ChrisJohnNOAA (approved):

Issue #139, opened 2024-03-21
Add a way for admins to disable/and or adjust the limits for the "Unusual activity" mail


ChrisJohnNOAA:
Request from Frederic Leclercq:

Add a way how to disable/and or adjust the limits for the "Unusual activity" mail.

The 25% is hard coded:
https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java#L1253

It'd be great to have the option to tweak that percentage or turn off those emails. Whenever there's a client or bot messing up requests, we end up flooded with those emails.

ayushsingh01042003:
I'd like to try and work on this.

Maybe we can add an option (something like a link within the email) that will send a request to the server to disable these emails for a certain duration.  

Can I get the steps to reproduce this email if possible?
- Where do I add the email who will get the spam?
- Is there anything specific I need to with the `datasets.xml` file?

ChrisJohnNOAA:
The email address is set in the setup.xml, for example if using the local Jetty server that is emailEverythingTo here:
https://github.com/ERDDAP/erddap/blob/main/development/jetty/config/setup.xml#L67

I think we should add a parameter to the datasets.xml that allows specifying the unusual activity threshold. Maybe something like unusualActivityFailPercent that defaults to 25 and has -1 as an option to disable the email.

The failure counter is incremented in Erddap.java: https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java#L747

To get the email, you should be able to just send requests to a local server that cause errors. You could set up a specific kind of request to throw an exception (temporarily to make testing easier).

PR #191, opened 2024-08-27
Feature: unusualActivityFailPercent

ayushsingh01042003:
## Description
- < unusualActivityFailPercent ></ unusualActivityFailPercent > tag for both SAX and simpleXml parsers in `datasets.xml`
- Any integer value from 0 to 100 to tweak the threshold percentage, Default = -1 for sending no mails 
- Fixes https://github.com/ERDDAP/erddap/issues/139

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
@@ -384,7 +384,7 @@ public class EDStatic {
   public static final int DEFAULT_slowDownTroubleMillis = 1000;
   public static final int DEFAULT_unusualActivity = 10000;
   public static final int DEFAULT_updateMaxEvents = 10;
-  public static final int DEFAULT_unusualActivityFailPercent = -1;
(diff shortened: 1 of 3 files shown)


ChrisJohnNOAA (approved):

ayushsingh01042003:
Have not really tested this. I think should work fine but let me know if otherwise

WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
```diff
@@ -384,6 +384,7 @@ public class EDStatic {
   public static final int DEFAULT_slowDownTroubleMillis = 1000;
   public static final int DEFAULT_unusualActivity = 10000;
   public static final int DEFAULT_updateMaxEvents = 10;
+  public static final int DEFAULT_unusualActivityFailPercent = -1;
```
ChrisJohnNOAA: Since this is a new parameter and existing installations will not have it set in their config, this should default to maintain the current behavior (25).

ChrisJohnNOAA (changes requested):

ChrisJohnNOAA (approved):

PR #198, opened 2024-09-11
Add prometheus jvm metrics. Also have erddap start on load (so datase…

ChrisJohnNOAA:
…t loading begins immediately).

To see the metrics on a running server the path is:

http://localhost:8080/erddap/metrics?debug=text

# Description

Adds basic Prometheus jvm metrics. This can be expanded on with additional metrics to track.

This also indicates to Tomcat/Jetty/others (in the web.xml) that the ERDDAP servlet should be loaded at startup which means dataset loading should now start immediately instead of waiting until the first request is made.

Fixes # (issue)

## Type of change

- [ X ] New feature (non-breaking change which adds functionality)
- [ X ] This change requires a documentation update

- [ X ] I have performed a self-review of my code
- [ X ] My code follows the style guidelines of this project
- [ X ] I have commented my code, particularly in hard-to-understand areas
- [ X ] I have made corresponding changes to the documentation
- [ X ] My changes generate no new warnings
- [ X ] I have added tests that prove my fix is effective or that my feature works
- [ X ] New and existing unit tests pass locally with my changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@ Simply run `mvn test` in a terminal to run the JUnit tests.
 
 Note that by default tests that do an image comparison are enabled. To disable those tests add `ImageComparison` to the `excludedGroups` section of the surefire `configuration`. It is recommended you run the image tests before making changes to ERDDAP&trade; so you can generate a baseline set of images that will be later used for comparison.
 
+### Metrics
+
+Metrics are collected using [Prometheus](https://prometheus.github.io/client_java/). You can see the metrics on a local server at [/erddap/metrics](http://localhost:8080/erddap/metrics?debug=text).
 
 ### Building a war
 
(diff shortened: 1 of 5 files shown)


srstsavage:
This is great! I'd imagine we probably need a setting to disable the metrics entirely?

ChrisJohnNOAA:
The current metrics I don't believe are sensitive. They are just general jvm data.

I added a setup.xml parameter, however just a note this doesn't stop the metrics servlet from running (the servlet is registered outside of java, in the web.xml). If prometheus metrics are disabled with the parameter that prevents the jvm metrics from getting registered. If we register other metrics without checking the parameter, they will be served.

If we want to use prometheus to track (and serve) the same metrics as shown on the status page, I'd prefer to only track them in one spot (prometheus). With the current configuration that would likely mean serving those specific metrics through prometheus even if the use prometheus parameter is off. Since those are already available publicly, maybe that's ok.

PR #203, opened 2024-09-23
Allow skipping resource download/unzip

srstsavage:
Allow skipping external resource download/unzip
using Maven properties.

`skipResourceDownload` skips downloading/unzipping erddapContent, ref, and etopo.

`skipTestResourceDownload` skips downloading/unzipping test data downloads data, largeFiles, largePoints, and largeSatellite.

Example invocation:

`mvn -DskipResourceDownload -DskipTestResourceDownload package`

These options can be also be set in the `MAVEN_ARGS` environment variable or in `.mvn/maven.config` (see https://maven.apache.org/configure.html).

# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

- [ ] I have performed a self-review of my code
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/PROGRAMMER_GUIDE.md
+++ b/PROGRAMMER_GUIDE.md
@@ -17,7 +17,10 @@ ERDDAP™ uses Maven to load code dependencies as well as some static reference 
 
   - [ref\_files.zip](https://github.com/ERDDAP/ERDDAPRefFiles/releases/download/1.0.0/ref_files.zip) and unzip it into /WEB-INF/ref/ .
 
-  - [erddapContent.zip](https://github.com/ERDDAP/erddap/releases/download/v2.23/erddapContent.zip) (version 2.23, 19810 bytes, MD5=1E26F62E7A06191EE6868C40B9A29362, dated 2023-02-27) and unzip it into _tomcat_, creating _tomcat_/content/erddap .
+  - [erddapContent.zip](https://github.com/ERDDAP/erddap/releases/download/v2.24/erddapContent.zip) (version 2.24, 19810 bytes, MD5=1E26F62E7A06191EE6868C40B9A29362, dated 2024-06-07) and unzip it into _tomcat_, creating _tomcat_/content/erddap .
+
+NOTE: Maven caches downloads but will unzip the downloaded archives on each execution, which takes time. To skip downloading
+and unzipping archives, you may specify the `skipResourceDownload` property to Maven (e.g. `mvn -DskipResourceDownload package`).
 
 - ERDDAP™ and its subcomponents have very liberal, open-source [licenses](https://erddap.github.io/setup.html#license), so you can use and modify the source code for any purpose, for-profit or not-for-profit. Note that ERDDAP™ and many subcomponents have licenses that require that you acknowledge the source of the code that you are using. See [Credits](https://erddap.github.io/setup.html#credits). Whether required or not, it is just good form to acknowledge all of these contributors.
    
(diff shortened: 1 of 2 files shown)


srstsavage:
When executing a single test (`mvn test -Dtest=LoadDatasetsTests`), skipping external resource downloads/unzips reduces the execution time from 2:42 minutes to 24.321 seconds (all of the saved time is due to skipped unzipping).

I also worked up a more elaborate working solution which moved the external resource unzipping out of `download-maven-plugin` to the `maven-antrun-plugin` in Maven profiles which activate when relevant files are missing (e.g. `content/erddap/setup.xml`, `WEB-INF/ref/etopo1_ice_g_i2.bin`, etc). This worked well, but I couldn't think of a satisfactory way to handle updates to the external resources; since the unzip profiles aren't run if the extracted target files exist, a developer would have to delete the target files or manually specify the unzip profiles to be activated in the Maven invocation to unzip the updated resource zips. It seemed much too easy for a developer to end up with out of date resource files in their project with this approach.

Using `skip` properties is straightforward and is safe by default (performing unzip on every execution and ensuring that new external resource files are propagated to the project directories) while still allowing significant speed ups to devs who knowingly opt into them by using `skipResourceDownload` and/or `skipTestResourceDownload`.

`skipTestResourceDownload` is a separate property for the specific use case of compiling the war in a fresh checkout without running tests.

srstsavage:
One more thought: it seems like packaging (at least some of) the external dependencies as Maven jar artifacts and hosting them as GitHub packages in a Maven registry might be a nice way to handle this. Then the dependencies could just be declared normally in the POM and used as-is or unpacked into place using `maven-dependency-plugin`'s `unpack`.

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry

https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages

> GitHub Packages usage is free for public packages.

https://maven.apache.org/plugins/maven-dependency-plugin/examples/unpacking-artifacts.html

Food for thought for a future approach maybe.

ChrisJohnNOAA:
> One more thought: it seems like packaging (at least some of) the external dependencies as Maven jar artifacts and hosting them as GitHub packages in a Maven registry might be a nice way to handle this. Then the dependencies could just be declared normally in the POM and used as-is or unpacked into place using `maven-dependency-plugin`'s `unpack`.
> 
> https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-apache-maven-registry
> 
> https://docs.github.com/en/billing/managing-billing-for-github-packages/about-billing-for-github-packages
> 
> > GitHub Packages usage is free for public packages.
> 
> https://maven.apache.org/plugins/maven-dependency-plugin/examples/unpacking-artifacts.html
> 
> Food for thought for a future approach maybe.

Based on my reading of the documentation, GitHub packages usage is free up to 1gb transfer (and 500mb of storage). We'd exceed that on just one sync of the resources. Based on my quick looking at the pricing, the costs could grow quickly ($0.25 per gb extra storage, and $0.50 per gb transfer).

Something else I've considered is adding the ability to avoid unzipping already downloaded resources to the maven-download-plugin: https://github.com/maven-download-plugin/maven-download-plugin

Though if I was going to put significant time into this problem, I'd like to reduce the size of the test resources we use. If the test resources were small enough, they could just be kept in the repo itself.

ChrisJohnNOAA (approved):

srstsavage:
> Something else I've considered is adding the ability to avoid unzipping already downloaded resources to the maven-download-plugin: https://github.com/maven-download-plugin/maven-download-plugin

Yeah, I had a similar thought, something like adding a `unzipWhenChanged` setting and checking for `!haveFile` here when it's set:

https://github.com/maven-download-plugin/maven-download-plugin/blob/master/src/main/java/com/googlecode/download/maven/plugin/internal/WGetMojo.java#L527

Seems straightforward, and it would propagate changed downloads to the project filesystem. It wouldn't defend against the developer getting the local external resources into a bad state though (deleting or changing extracted files, etc). Maybe that's sufficient?

> Based on my reading of the documentation, GitHub packages usage is free up to 1gb transfer (and 500mb of storage).

Their documentation seems a big ambiguous to me on whether those limits only apply to private packages. If I get some time I'll do some experimentation outside of the ERDDAP org/repo to find out.

ChrisJohnNOAA:
Awesome. Hopefully they'll accept it.

PR #219, opened 2024-10-17
Release 2.25

ChrisJohnNOAA:
# Description

Changes for the 2.25 release including updating version numbers, some documentation updates, excluding some unnecessary files from the war, and defaulting the SharedWatchService to on.

## Type of change


srstsavage:
@ChrisJohnNOAA For `download-maven-plugin`, I'm noticing some odd behavior when using `unpackWhenChanged` (unextracted zip files remain in the output directory on every other run) that I'd like to fix before setting `unpackWhenChanged` to true. I think it still makes sense to upgrade `download-maven-plugin` to 1.11.0 so download progress can be quieted with `--no-transfer-progress`, but my vote would be to hold off on `unpackWhenChanged` until that behavior is fixed.

ChrisJohnNOAA:
> @ChrisJohnNOAA For `download-maven-plugin`, I'm noticing some odd behavior when using `unpackWhenChanged` (unextracted zip files remain in the output directory on every other run) that I'd like to fix before setting `unpackWhenChanged` to true. I think it still makes sense to upgrade `download-maven-plugin` to 1.11.0 so download progress can be quieted with `--no-transfer-progress`, but my vote would be to hold off on `unpackWhenChanged` until that behavior is fixed.

Sure. I switched it back to unpack.

Issue #166, opened 2024-06-28
Code uses emailDailyReportTo, documentation references emailDailyReportsTo


srstsavage:
For the emailDailyReportTo setting, there is a mismatch between the EDStatic code reading setup.xml or environment variables (which references `emailDailyReportTo`) and the setup.xml documentation (which references `emailDailyReportsTo`).

EDStatic code reading `emailDailyReportTo`:

https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java#L1843

Various documentation/comments referencing `emailDailyReportsTo`:

https://github.com/search?q=repo%3AERDDAP%2Ferddap%20emailDailyReportsTo&type=code
https://github.com/ERDDAP/erddapContent/blob/main/content/erddap/setup.xml#L56
https://github.com/ERDDAP/erddapContent/blob/main/content/erddap/setup.xml#L69

These should be aligned, or EDStatic should be adjusted to accept both the singular and plural.

ChrisJohnNOAA:
I think making EDStatic accept both would be the better solution.

ayushsingh01042003:
To confirm if I understand this:

In setup.xml the tag is `emailDailyReportsTo` but EDStatic is trying to get `emailDailyReportTo` instead, which I assume is a typo?
If that is the case then in this code here
```
  private static String getSetupEVString(
      ResourceBundle2 setup, Map<String, String> ev, String paramName, String tDefault) {
    String value = ev.get("ERDDAP_" + paramName);
    if (String2.isSomething(value)) {
      String2.log("got " + paramName + " from ERDDAP_" + paramName);
      return value;
    }
    return setup.getString(paramName, tDefault);
  }
```

`setup.getString()` would always return null since `emailDailyReportTo` does not exist in setup.xml file and only the `ev.get()` method is working.

Why don't we just update it to `emailDailyReportsTo` in this line
```
emailDailyReportToCsv =
          getSetupEVString(setup, ev, "emailDailyReportTo", ""); // won't be null
```
is it important to support both `emailDailyReportsTo` and `emailDailyReportTo`?

ChrisJohnNOAA:
It would be good to support both emailDailyReportsTo and emailDailyReportTo. Users have their own setup.xml and some likely followed the documentation and set emailDailyReportsTo while some may have figured out the proper name is emailDailyReportTo.

[...]

PR #236, opened 2024-12-11
Optional check for emailDailyReportsTo in setup.xml

ayushsingh01042003:
Issue [#116](https://github.com/ERDDAP/erddap/issues/166)
Checks if the setup.xml has `emailDailyReportTo` if not it checks for `emailDailyReportsTo`

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
@@ -84,6 +84,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
```diff
@@ -1933,7 +1934,8 @@ public static int convertToPublicSourceUrlFromSlashPo(String tFrom) {
       emailFromAddress = getSetupEVString(setup, ev, "emailFromAddress", (String) null);
       emailEverythingToCsv = getSetupEVString(setup, ev, "emailEverythingTo", ""); // won't be null
       emailDailyReportToCsv =
-          getSetupEVString(setup, ev, "emailDailyReportTo", ""); // won't be null
+          Optional.ofNullable(getSetupEVString(setup, ev, "emailDailyReportTo", ""))
```
ChrisJohnNOAA: Because this defaults to a string (""), if there is no value for emailDailyReportTo in setup.xml this will just use "".

I believe this should work:
```
emailDailyReportToCsv =
          Optional.ofNullable(getSetupEVString(setup, ev, "emailDailyReportTo", (String) null))
              .orElse(getSetupEVString(setup, ev, "emailDailyReportsTo", ""));
```

ChrisJohnNOAA (changes requested):
The first read needs to default to null not "".

WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
ayushsingh01042003: Thanks for pointing that out

ChrisJohnNOAA (approved):
This looks good. It can be merged once the tests pass.

ChrisJohnNOAA:
It seems the watch directory tests require EDStatic initialization. For whatever reason it seems the test execution order changed. I made a small change to fix the issue.

Issue #134, opened 2024-02-23
Dataset Citation and display info


ChrisJohnNOAA:
From Bob Simons:

> In datasets.xml, there could be an optional tag like
>   <displayInfo>attribute1,display1,attribute2,display2</displayInfo>
> which can have 1+ pairs of values (a global attribute name and the word to display it as on web pages,
> which tells ERDDAP: if attribute1 exists for a given dataset, display the display1 word and an (?) icon (which reveals the attribute value).
> For Claudia, this could be
>   <displayInfo>Datacite,Citation</displayInfo>
> which would tell ERDDAP: if the dataset has a "Datacite" global attribute, display it in the dataset information section on web pages as:  
>    Citation (?)
> where hovering over the (?) icon shows the Datacite value.
> The good thing about this approach is that it gives the administrator full control over exactly how the citation appears. If one group likes the Datacite format, great. If another group likes some other format, that's fine for them.
> I hope that make sense.
> I think this is very useful to a wide variety of groups. It is very flexible. I think it is not hard to implement.

For additional context this was discussed in this thread on the users group:
https://groups.google.com/g/erddap/c/k58VhLE7MhA/m/o8dZvohIAwAJ?utm_medium=email&utm_source=footer

ayushsingh01042003:
This can be closed with the merge of #240

ChrisJohnNOAA:
Yes. Thank you for the contribution!

PR #240, opened 2025-01-14
displayInfo and displayAttribute tags for datasets.xml

ayushsingh01042003:
Issue https://github.com/ERDDAP/erddap/issues/134

- Tags `<displayInfo>` and `<displayAttribute>` are now available as global level tags in `datasets.xml`
- They should contain the same number of elements seperated by a `,`
- `displayAttribute` is the attribute name from the top level `addAttributes` tag of a dataset whose value is displayed in the tooltip
- `displayInfo` should contain what the word should be beside the tooltip
- If tags are not added in `datasets.xml` the default values are "summary" and "license" which work the same

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
@@ -229,12 +229,7 @@ public class LoadDatasets extends Thread {
       int nTry = nTryAndDatasets[0];
       int nDatasets = nTryAndDatasets[1];
 
-      // validation for displayInfo and displayAttribute Tags
-      if (EDStatic.displayAttributeAr.length != EDStatic.displayInfoAr.length) {
-        String2.log("Incorrect input to the displayAttribute and displayInfo tags");
-        EDStatic.displayAttributeAr = EDStatic.DEFAULT_displayAttributeAr;
-        EDStatic.displayInfoAr = EDStatic.DEFAULT_displayInfoAr;
-      }
+      validateDatasetsXmlResults();
 
       erddap.updateLucene(changedDatasetIDs);
       lastLuceneUpdate = System.currentTimeMillis();
@@ -453,6 +448,15 @@ public class LoadDatasets extends Thread {
       synchronized (EDStatic.suggestAddFillValueCSV) {
         EDStatic.suggestAddFillValueCSV.setLength(0);
       }
+    }
+  }
+
+  /** Validation checks post the parsing of the datasets.xml file */
+  private void validateDatasetsXmlResults() {
+    if (EDStatic.displayAttributeAr.length != EDStatic.displayInfoAr.length) {
+      String2.log("Incorrect input to the displayAttribute and displayInfo tags");
+      EDStatic.displayAttributeAr = EDStatic.DEFAULT_displayAttributeAr;
+      EDStatic.displayInfoAr = EDStatic.DEFAULT_displayInfoAr;
     }
   }
 
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
@@ -229,6 +229,13 @@ public class LoadDatasets extends Thread {
       int nTry = nTryAndDatasets[0];
       int nDatasets = nTryAndDatasets[1];
 
+      // validation for displayInfo and displayAttribute Tags
+      if (EDStatic.displayAttributeAr.length != EDStatic.displayInfoAr.length) {
+        String2.log("Incorrect input to the displayAttribute and displayInfo tags");
+        EDStatic.displayAttributeAr = EDStatic.DEFAULT_displayAttributeAr;
+        EDStatic.displayInfoAr = EDStatic.DEFAULT_displayInfoAr;
+      }
+
       erddap.updateLucene(changedDatasetIDs);
       lastLuceneUpdate = System.currentTimeMillis();
 
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDD.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDD.java
@@ -3795,63 +3795,73 @@ public abstract class EDD {
             + "</table>\n");
   }
 
+  /**
+   * This renders the string for the Information section of the dataset page
+   *
+   * @return the String to append to Information row
+   */
   private String getDisplayInfo(int language, String loggedInAs) {
     if (EDStatic.displayAttributeAr.length != EDStatic.displayInfoAr.length) {
+      String2.log("Incorrect input to the displayAttribute and displayInfo tags");
       return "";
     }
 
     StringBuilder displayInfoStr = new StringBuilder();
-    String tSummary = extendedSummary();
-    displayInfoStr
-        .append(EDStatic.EDDSummaryAr[language])
-        .append(" ")
-        .append(
+
+    for (int i = 0; i < EDStatic.displayAttributeAr.length; i++) {
+      String attribute = EDStatic.displayAttributeAr[i];
+      String displayInfo = EDStatic.displayInfoAr[i];
+      String value;
+
+      // Handle "summary"
+      if ("summary".equals(attribute)) {
+        String tSummary = extendedSummary();
+        value =
             EDStatic.htmlTooltipImage(
                 language,
                 loggedInAs,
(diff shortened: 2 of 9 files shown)


ayushsingh01042003:
TODO: 
- I had some difficulty in understanding the `edd.ensureValid()`  hence it is just an if statement for now that checks if lengths of the array is the same. It means no error is given if tags have different lengths and summary and license are handled as usual.

- Code structuring may be a little random for this PR but it seems to be working correctly:

Screenshot from 2025-01-14 13-53-44

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDD.java
```diff
@@ -3820,6 +3821,39 @@ public void writeHtmlDatasetInfo(
...
+    if (EDStatic.displayAttributeAr.length != EDStatic.displayInfoAr.length) {
+      return "";
+    }
+
+    StringBuilder displayInfoStr = new StringBuilder();
+    for (int i = 0; i < EDStatic.displayAttributeAr.length; i++) {
+      String attribute = EDStatic.displayAttributeAr[i];
+
+      // Skip "summary" and "license" as they are already handled
+      if ("summary".equals(attribute) || "license".equals(attribute)) {
```
ChrisJohnNOAA: Summary and license should be included in this general version. That does mean removing their code above.

It'd be ideal to have their display attribute localized (using the EDStatic.EDDSummmaryAr[language] for summary and similar for license).

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDD.java
```diff
@@ -3820,6 +3821,39 @@ public void writeHtmlDatasetInfo(
             + "</table>\n");
   }
 
+  private String getDisplayInfo(int language, String loggedInAs) {
+    if (EDStatic.displayAttributeAr.length != EDStatic.displayInfoAr.length) {
```
ChrisJohnNOAA, line 3804: A good spot to do the validation would be in LoadDatasets.java just after the parsing is complete:
https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java#L229

If the validation fails, it should log the issue and then set the EDStatic.displayAttributeAr and EDStatic.displayInfoAr to the defaults.

It's probably worth keeping a check here to make sure they match just before using them.

ChrisJohnNOAA (changes requested):
This is a good start. I have a couple of comments overall:

It'd be good to add tests. In particular a test that the parameters are parsed correctly and a test that they are used correctly. TopLevelHandlerTests.java would likely be a good spot for the parsing test (do note the comment about making sure EDStatic is not modified for other tests, so for now that means setting these values back to default when the test is complete). As for testing the usage, most likely that will be best done in JettyTests.java. You'll probably need to modify EDDTestDataset.generateDatasetsXml to include the new top level parameters and then in a new test in the JettyTests.java file, use SSR.getUrlResponseStringUnchanged to load a page for a dataset (that has the proper attributes) and make sure the new attributes are properly rendered.

This is currently only for the SAX parser and currently requires setting useSaxParser to true as well. That's fine but that should definitely be documented somewhere along with how to use it. Would you be able to include a short markdown file explaining the feature and that it's SAX parser only?

A couple of comments in the code as well.

ChrisJohnNOAA:
> I made some progress but for the jetty tests when I run the command `mvn -Dtest=JettyTests test` the output is always Tests run: 0, Failures: 0, Errors: 0, Skipped: 0 Is there some other way to run the Jetty Tests?

The JettyTests are part of the integration tests, You can run them with mvn verify, though that will run all of the integration tests. You should be able to run:
```
mvn -D it.test=JettyTests verify
```
or
```
mvn -Dtest=JettyTests verify
```
to run just the JettyTests I've seen both in documentation, haven't verified which one works myself.
You can most likely run an individual JettyTest (in any command that's working with JettyTests#specificTestName.

ayushsingh01042003:
So I tried using the above commands with '#testName'  to run an individual integration test but the problem is it runs all the unit tests as well but when I run it with `-DskipTests` flag it skips both unit and integration tests and runs nothing.
Could you confirm the command for running a singular integration test?

WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
```diff
@@ -229,6 +229,13 @@ public void run() {
       int nTry = nTryAndDatasets[0];
       int nDatasets = nTryAndDatasets[1];
 
+      // validation for displayInfo and displayAttribute Tags
```
ChrisJohnNOAA: Can we put the validation in a function (validateDatasetsXmlResults  or something). I know it's only one check right now, but making it a function would help make a place for future validation to be placed.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDD.java
ChrisJohnNOAA: There should be a log here as well.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDD.java
ChrisJohnNOAA: I meant that summary and license should be included in this loop.

ChrisJohnNOAA (changes requested):
I haven't run an individual integration test, so if those commands from the documentation are not working I'd need to look into it.

Adding to SimpleXmlParser logic would be good, though not needed, I'm happy with this being a feature to try to encourage people to update to the SAX parrser.

Yes, a file with.md syntax. I'm in the process of migrating all of the documentation to markdown format so that we can use Docusaurus for our documentation site.

Code code is on the right track, I've added some comments.

ayushsingh01042003:
I am not sure how I can add summary and license to the loop. Lets say an if condition for summary and another for license but since these two fields always need to be added, if the user enters
<displayAttribute>att1,license,att3</displayAttribute>
We will end up not appending summary at all in the UI since if(attribute == "summary") block never runs

Was there some other way you were referring to?

ayushsingh01042003:
## `displayInfo` and `displayAttribute` Tags

### Description
This feature allows you to display global attributes of your choice on the datasets page in the `Information` row.

### Usage Instructions
These tags can only be used with the `Sax parser`. To enable and use them, follow these steps:

1. **Enable the SAX Parser**:
   Add the following line to your `setup.xml` file:
   ```xml
   <useSaxParser>true</useSaxParser>
   ```

2. **Add Tags in `datasets.xml`**:
   In the `datasets.xml` file, include two top-level tags:
   ```xml
   <displayInfo></displayInfo>
   <displayAttribute></displayAttribute>
   ```

3. **Default Behavior**:
   - If these tags are not added or left empty in the `datasets.xml` file, the default values are applied as follows:
     - `displayInfo`: `Summary,License`
     - `displayAttribute`: `summary,license`

4. **Ensure Consistency**:
   The number of comma-separated values in both `displayInfo` and `displayAttribute` tags must be the same.

### How It Works
- The `displayAttribute` tag specifies global attributes (defined within the `<addAttributes>` tag) to be displayed for each dataset.
- The corresponding values in the `displayInfo` tag are displayed as labels in the `Information` row of the UI.
- When the user hovers over the displayed labels, a tooltip will appear, showing the value of the global attribute.

### Example
```xml
<displayInfo>Display1,Display2</displayInfo>
<displayAttribute>att1,att2</displayAttribute>
```

#### Dataset Global Attributes Example:
```xml
<att name="att1">This is att1</att>
<att name="att2">This is att2</att>
```

#### UI Behavior:
- The words `Display1` and `Display2` will be displayed in the `Information` row on the UI.
- When hovered, tooltips will display the corresponding attribute values:
  - `Display1`: Tooltip shows _This is att1_
  - `Display2`: Tooltip shows _This is att2_

### Notes
- Ensure the attribute names specified in the `displayAttribute` tag match the global attributes defined in the dataset.
- Incorrect or missing attributes will log error messages.

By following these steps, you can customize the `Information` row on the datasets page to display relevant global attributes with corresponding tooltips.

ayushsingh01042003:
- Hope the above comment works for the markdown file
- Let me know about any comments and changes to the markdown file or the code contributions

ChrisJohnNOAA:
> * Hope the above comment works for the markdown file
> * Let me know about any comments and changes to the markdown file or the code contributions

The documentation is great, Thanks for writing that!

I've sent a [pull request](https://github.com/ayushsingh01042003/erddap/pull/4) that simplifies the code a little bit and fixes one of the JettyTests. Let me know if you have questions on those changes.

ChrisJohnNOAA (approved):

Issue #163, opened 2024-06-26
Add aspect ratio as a parameter for image requests


ChrisJohnNOAA:
&.yRange=min|max|ascending|scale|**aspect**

aspect would be a decimal between 0.01 and 100

originally requested in email by John Kerfoot.

I believe this requires modifying EDDGrid and EDDTable saveAsImage method (in both) to parse the new value.

yousifelboukhary:
@ChrisJohnNOAA 
I opened [#243](https://github.com/ERDDAP/erddap/pull/243) this PR.

PR #243, opened 2025-01-20
add aspect ratio as a parameter for image requests

yousifelboukhary:
# Description

I modified EDDGrid and EDDTable saveAsImage method (in both) to parse the new value.

Fixes [#163](https://github.com/ERDDAP/erddap/issues/163)

> aspect would be a decimal between 0.01 and 100
For this, when the aspect value goes out of bounds, I set it back to Double.NaN indicating the aspect is not set. 

- [ ] My code follows the style guidelines of this project

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGrid.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGrid.java
@@ -7130,6 +7130,7 @@ public abstract class EDDGrid extends EDD {
       double minX = Double.NaN, maxX = Double.NaN, minY = Double.NaN, maxY = Double.NaN;
       boolean xAscending = true, yAscending = true; // this is what controls flipping of the axes
       String xScale = "", yScale = ""; // (default) or Linear or Log
+      double aspect = Double.NaN;
       int nVars = 4;
       EDV vars[] = null; // set by .vars or lower
       int axisVarI[] = null, dataVarI[] = null; // set by .vars or lower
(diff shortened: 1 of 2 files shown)


ChrisJohnNOAA (changes requested):
I appreciate you making the pull request. Unless I'm missing something this doesn't solve the issue though. This does parse the parameter which is a great start, but it doesn't use that value. I believe you need to set imageHeight = imageWidth * aspect or something similar. Also, we only want this to impact images (not pdf requests), so before adjusting the imageHeight, that should be checked.

Also I'm trying to make sure there is good unit test coverage of ERDDAP code. Please write a unit test that covers the new contributions.

ChrisJohnNOAA:
> @ChrisJohnNOAA Thank you for the clarifications! I will work on it. Is there any documentation about the algorithm implemented that would help me understand the function more, or any relevant information that would be appreciated?

Much of the saveAsImage code for both EDDTable and EDDGrid is preparing parameters to call one (or more) of the main functions that do drawing:

SgtMap.makeMap
https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/coastwatch/sgt/SgtMap.java#L451

SgtMap.makeCleanMap
https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/coastwatch/sgt/SgtMap.java#L1992

SgtGraph.makeGraph
https://github.com/ERDDAP/erddap/blob/main/WEB-INF/classes/gov/noaa/pfel/coastwatch/sgt/SgtGraph.java#L136

I'm not sure exactly what your question is, but it could be useful to read their Javadoc comments to understand how the variables in saveAsImage are used.

PR #274, opened 2025-04-17
Download file type refactor and semantic versioning

ChrisJohnNOAA:
# Description

This refactors the code for file downloads to make it easier to add a new file type to ERDDAP. Because the file types were the primary space that ERDDAP internally used the version, I also updated the code to support semantic versioning while I was modifying it all.

Fixes # (issue)

## Type of change


clifford-harms:
This is a big deal, much needed change in my opinion.  This is one of the key elements (along with similar work for the dataset ingest) that will allow extending ERDDAP formats without fiddling with Erddap.java.  I like the cache integration, migration of existing outputs to the new interfaces, and encapsulation of request info (DapRequestInfo), so no more  6+ argument methods.  I may have missed it (large PR), does FileTypeClass have everything that is needed to expose the output format in the existing GUI?

ChrisJohnNOAA:
> This is a big deal, much needed change in my opinion. This is one of the key elements (along with similar work for the dataset ingest) that will allow extending ERDDAP formats without fiddling with Erddap.java. I like the cache integration, migration of existing outputs to the new interfaces, and encapsulation of request info (DapRequestInfo), so no more 6+ argument methods. I may have missed it (large PR), does FileTypeClass have everything that is needed to expose the output format in the existing GUI?

Thanks!

That is the goal of FileTypeClass, however the localized descriptions were complicated. The way the annotation information is used is through a helper in EDD: EDDFileTypeInfo. This also hides from the rest of the code that for localized descriptions we instantiate the class and get the localized text that way.

A new filetype needs to be implement FileTypeInterface and have the FileTypeClass annotation.

Issue #96, opened 2023-03-16
Relative URLs


jwithamfugro:
Similar request to #27 

We have an ERDDAP instance running on a private network machine which cannot be added to DNS. Because we want this machine to have a domain, we've set the baseUrl to the domain we want to use anyway (e.g. "`http://erddap.localdomain:8080`") and ask each user to modify their LMHOSTS and HOSTS file so their browsers recognise it and can properly navigate ERDDAP.

Obviously, there are issues when someone on the network who doesn't have this localdomain setup tries to navigate ERDDAP from direct IP, as webpage links use the baseUrl. With DNS out of the question, we are considering if it's possible for links on webpages to use relative URLs rather than baseUrl, so that it doesn't matter what name you are using to access ERDDAP. Link sharing isn't that much of a concern, as contradictory as it may sound.

We tried #27's little hack and made the baseUrl "`/.`", and while that does make webpage links resolve to the domain the user is accessing ERDDAP from, it's an incorrect configuration that causes issues with at least one other part of ERDDAP; namely the URL generation on tabledaps treating the baseUrl as literal and outputting URLs which cannot be used by anyone (eg: "`/./erddap/tabledap/datasetname.htmlTable`"). To avoid issues like that, we would rather prefer the ability to make webpages respect the domain the user is accessing from regardless of what the baseUrl is.

srstsavage:
+1, my vote would be a boolean setting to enable ERDDAP to use the values of `X-Forwarded-Host`, `X-Forwarded-Port`, and `X-Forwarded-Proto` to build response URLs, fall back to the `Host` header if not present, and fall back again to the `baseUrl` setting if no headers are present.

BobSimons:
I thought of one of the situations where absolute URLs are an advantage: in a grid of ERDDAPs [https://coastwatch.pfeg.noaa.gov/erddap/download/grids.html#grids](https://coastwatch.pfeg.noaa.gov/erddap/download/grids.html#grids), this allows the secondary (not officially public) ERDDAPs to push users back to the main, public ERDDAP. 
I think there are other cases where absolute urls are useful, but can't remember.

[...]

PR #290, opened 2025-05-18
Support useHeadersForUrl feature flag (use Host header when building urls)

srstsavage:
# Description

Add support for a useHeadersForUrl feature flag, which changes EDStatic methods erddapUrl() and erddapHttpsUrl() to use request `Host` header when building urls if useHeadersForUrl is true and the request is available/not null.

Functionally, when this flag is enabled ERDDAP can accessed on any address which resolves to it. For example an ERDDAP available at https://erddap.officialdomain.org/erddap, http://server.localdomain:8080/erddap, and http://localhost:8080/erddap will generate URLs consistent with the accessed hostname.

This change makes the assumption that most URLs written in response to user interaction (e.g. loading a page in a browser) should respect the Host header in the request, but some URLs (example RSS feeds) should continue to always use the erddapPreferredUrl, which is constructed from baseUrl and baseHttpsUrl uses baseHttpsUrl if available, or baseUrl if not). If the header driven URL constructing approach becomes the default in the future, we could imagine only having to supply a single ERDDAP preferred URL config to use when constructing URLs outside of a request context (generated emails, etc) instead of separate baseUrl and baseHttpsUrl.

One corner case is the behavior when a user accesses a login page over http. When useHeadersForUrl is enabled and EDStatic.erddapHttpsUrl is called with an http request, the resulting URL prefix is upgraded to https with the request provided Host header as long as there is not a port in the request Host (i.e. a request for http://erddap.domain.org would result in a https://erddap.domain.org base URL, but a request for http://erddap.domain.org:8080 would fall back to returning the erddapHttpsUrl instead since https://erddap.domain.org:8080 is nonsensical).

Absolute URLs are constructed from the request Host header instead of path-only ("/erddap/index.html") or relative ("../status.html") URLs. Returning path-only URLs would be technically simple, and I originally included an option in EDStatic.erddapUrl() to generate them, but removed it for simplicity. EDStatic.erddapUrl() is called a vast number of times, and the risk of bug introduction by generation of path-only or relative URLs (non-interactive contexts etc) seemed to justify keeping the URLs absolute, at least for this change. As such, this change doesn't implement relative URLs but does address many of the concerns in #96.

General expectations of behavior:

If useHeadersForUrl is false, erddapUrl or erddapHttpsUrl are used (legacy behavior).

If the request is null, erddapUrl or erddapHttpsUrl are used (legacy behavior).

If useHeadersForUrl is true and request is not null, scheme and host header from request are used.

If useHeadersForUrl true, request not null, and request scheme is http, EDStatic.erddapHttpsUrl() uses https scheme with request host as long as request host does not contain a port (i.e. http url are upgraded to https with the same host). If host does contain a port, erddapHttpsUrl is returned instead.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/coastwatch/util/FileVisitorDNLS.java
+++ b/WEB-INF/classes/gov/noaa/pfel/coastwatch/util/FileVisitorDNLS.java
@@ -1142,8 +1142,8 @@ public class FileVisitorDNLS extends SimpleFileVisitor<Path> {
    *
    * @param tDir The starting directory, with \\ or /, with or without trailing slash, which will be
    *     removed.
-   * @param startOfUrl usually EDStatic.erddapUrl(loggedInAs, language) + "/files/" + datasetID() +
-   *     "/" which will be prepended.
+   * @param startOfUrl usually EDStatic.erddapUrl(request, loggedInAs, language) + "/files/" +
+   *     datasetID() + "/" which will be prepended.
    * @return a table with columns with DIRECTORY (always "/"), NAME, LASTMODIFIED (long
    *     epochMilliseconds), and SIZE (long) columns.
    */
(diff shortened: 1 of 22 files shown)


srstsavage:
Also totally flexible on the feature flag name (currently `useHeadersForUrl`) if its not sufficiently clear

ChrisJohnNOAA (approved):

ChrisJohnNOAA:
Documenting what admins need to know would be appreciated. Most likely that would be adding a new md file or a section to one of the existing files here: https://github.com/ERDDAP/erddap.github.io/tree/main/docs/server-admin

Issue #295, opened 2025-05-30
Use JTE templating framework for rendering HTML


srstsavage:
Currently ERDDAP HTML is generated via manual string concatenation. Use of a templating engine to render HTML would simplify HTML maintenance/management, lower the learning curve for contributors to update UI HTML, provide a modern foundation for UI improvements, and potentially allow more UI customization options at deploy time.

[JTE](https://jte.gg/) has been identified as an optimal template engine due to its active development and focus on performance via template precompilation.
 
@Harish-2003 will be working on this project this summer (2025) via GSoC!

Associated Google Summer of Code project: https://summerofcode.withgoogle.com/programs/2025/projects/WFMQYGxF

GSoC project description:

> This project aims to modernize ERDDAP™ page rendering by migrating it to a templating framework. The current rendering process relies on manually constructed HTML, which makes it difficult to maintain and customize. By implementing a template-based system, we will improve code maintainability, enhance flexibility for customization, and support modern web development practices. Deliverables Identify a suitable templating framework (e.g., FreeMarker, Thymeleaf, or another appropriate choice). Refactor ERDDAP™’s HTML generation logic to use the selected template engine. Ensure the new system maintains full feature parity with existing pages. Improve documentation on how to modify and extend ERDDAP™ templates. Implement tests to verify rendering consistency. This upgrade will make ERDDAP™ more maintainable and adaptable for future development.

PR #301, opened 2025-06-07
Add JTE template dependencies, config, and initial templates

Harish-2003:
# Addition of JTE

Added the dependencies for the jte_templates ([JTE](https://jte.gg/)) in the pom.xml file,
added the feature flag to enable and disable the jte_templates
added an basic template for the status page 

working on #295 

## Feature


Harish-2003:
> > need to check the status.jte page as we can't able to do the html thing while passing like below i was getting plain text in the code rather than the html output need help with that
> > ```
> > @raw${model.youAreHere}@endraw
> > <pre>${model.statistics}</pre>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > also tried this
> > ```
> > @raw<div>${model.youAreHere}</div>@endraw
> > <pre>${model.statistics}</pre>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > also tried using @var none of them works
> 
> Have you tried using @unsafe https://jte.gg/html-rendering/#unsafe
> 
> Also, rather than rending the youAreHere to string and passing that through the model, I think we'd like to be able to have a youAreHere template, and call that from within the page. https://jte.gg/syntax/#template-calls

made the changes as you suggested please let me know if i need to do anything better @ChrisJohnNOAA  @srstsavage

src/main/jte/youarehere.jte
```diff
@@ -0,0 +1,4 @@
+@import gov.noaa.pfel.erddap.util.TemplateModel
+@param String youAreHere
+
+<div>$unsafe{youAreHere}</div>
```
ChrisJohnNOAA: Ideally this would take the parameters needed to build the youAreHere html and generate it in the template instead of in the Java and passing the resulting string to here.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -4997,24 +4994,35 @@ public void doStatus(
...
-      writer.write(XML.encodeAsHTML(sb.toString()));
-      writer.write("</pre>\n");
+
+      String statisticsHtml = XML.encodeAsHTML(sb.toString());
+
+      writer.write("<div class=\"standard_width\">\n");
+
+      if (EDStatic.config.useHtmlTemplates) {
+        // Use JTE template engine to render the status page
+        CodeResolver codeResolver =
```
srstsavage: I could be missing something, but `codeResolver` doesn't seem to be used and wasn't needed when I tested locally. Referencing `${project.basedir}/src/main/resources/jte/` won't work outside of the project source directory also, but I don't think we need to solve that problem since we're just using the precompiled template classes in the classpath. I think this line can just be removed.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -4997,24 +4994,35 @@ public void doStatus(
...
+            new DirectoryCodeResolver(Path.of("${project.basedir}/src/main/resources/jte/"));
+        TemplateEngine engine = TemplateEngine.createPrecompiled(ContentType.Html);
+        TemplateOutput jteOutput = new StringOutput();
+        TemplateModel model = new TemplateModel(youAreHereHtml, statisticsHtml);
+        engine.render("status.jte", Map.of("model", model), jteOutput);
+        writer.write(jteOutput.toString());
```
srstsavage: It works to use `gg.jte.output.StringOutput` and `writer.write(jteOutput.toString());`, but `gg.jte.output.WriterOutput` also works and would be more efficient (no need to hold the rendered template string in memory), so my vote would be to use that.

```
WriterOutput jteOutput = new WriterOutput(writer);
engine.render("status.jte", Map.of("model", model), jteOutput);
```

or just

```
engine.render("status.jte", Map.of("model", model), new WriterOutput(writer));
```

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -4965,28 +4972,18 @@ public void doStatus(
...
-      writer.write(
-          "<div class=\"standard_width\">\n"
-              + EDStatic.youAreHere(
-                  request, language, loggedInAs, EDStatic.messages.statusAr[language])
-              + "<pre>");
+      String youAreHereHtml =
+          EDStatic.youAreHere(request, language, loggedInAs, EDStatic.messages.statusAr[language]);
+
       StringBuilder sb = new StringBuilder();
       EDStatic.addIntroStatistics(sb, EDStatic.config.showLoadErrorsOnStatusPage, this);
```
srstsavage: Similar to @ChrisJohnNOAA's comment about the `youAreHere` template, instead of building the response Html using StringBuilder in `EDStatic.addIntroStatistics` and rendering that concatenated string in a thin JTE template, we want to build up a model object containing all of the data properties needed to build the status response (all the properties used in `EDStatic.addIntroStatistics` including currentTime, startupTime, loadTime, timeSinceLoad, currentlyLoadingDataset, numGridDatasets, numUniqueUsers, etc etc...) and then pass that status data model object to the status JTE template where the detailed HTML is built. It's a big job to build that data model class, but it will be much cleaner and give us a lot of flexibility and clarify on how the HTML is rendered!

Harish-2003: so you are saying that building two things one for jte and the one for string builder(which is already existed)? as we are currently using featureflag option?

Harish-2003: so without changing it i would rather be adding that in the if block of jte feature flag on function?

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
Harish-2003: ya checked it and its mentioned it perfectly fine without the requirment of the line, i thought it might be necessary as stated in the documentation 
removing the line as it works completely fine

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
Harish-2003: just made the changes as you suggested shane

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
Harish-2003: okay got it shane , correct me if i am wrong , i will try build a data model object containing the properties and then try to pass those values into the template using engine.render(,mapof(),) leaving in place of current values as they will help when featureflag is off.

srstsavage: That's basically it, yeah. Leave `EDStatic.addIntroStatistics` as-is because it will be used when feature flag `useHtmlTemplates` is false, but use it as a guide for re-creating the status page output in the JTE template.

Here are some examples I experimented with today. Since all of the status data seems to be static, the JTE status model class can have static methods as well. This `Status.java` JTE model class provides access to the first few properties used in `EDStatic.addIntroStatistics`, but doesn't build and HTML and only exists to provide these properties to the JTE template.

```
package gov.noaa.pfel.erddap.jte;

import com.cohort.util.Calendar2;
import gov.noaa.pfel.erddap.util.EDStatic;

public class Status {
    public static String getCurrentTimeLocal() {
        return Calendar2.getCurrentISODateTimeStringLocalTZ();
    }

    public static String getStartupTimeLocal() {
        return EDStatic.startupLocalDateTime;
    }

    public static String getTimeSinceLastMajorLoadDatasetsStarted() {
        return Calendar2.elapsedTimeString(
                System.currentTimeMillis() - EDStatic.lastMajorLoadDatasetsStartTimeMillis);
    }

    public static long getLoadDatasetsElapsedTimeSeconds() {
        return (EDStatic.lastMajorLoadDatasetsStopTimeMillis - EDStatic.lastMajorLoadDatasetsStartTimeMillis) / 1000;
    }

    public static boolean isLoadDatasetsRunning() {
        // load datasets stop time > start time means the current load hasn't finished
        return getLoadDatasetsElapsedTimeSeconds() < 0;
    }
}
```

And here is a simple `YouAreHere` JTE model class (only handles the simplest case with a single protocol, will need to be expanded to handle the more complex cases with help links etc).

```
package gov.noaa.pfel.erddap.jte;

import gov.noaa.pfel.erddap.util.EDStatic;

public class YouAreHere {
    private final String erddapUrl;
    private final String erddapLinkTitle;
    private final String protocol;
 
    public YouAreHere(final String erddapUrl, final String erddapLinkTitle, final String protocol) {
        this.erddapUrl = erddapUrl;
        this.erddapLinkTitle = erddapLinkTitle;
        this.protocol = protocol;
    }

    public String getErddapUrl() {
        return erddapUrl;
    }

    public String getErddapLinkTitle() {
        return erddapLinkTitle;
    }

    public String getProgramName() {
        return EDStatic.ProgramName;
    }

    public String getProtocol() {
        return protocol;
    }
}
```

And then the (incomplete!) JTE templates:

`status.jte`

```
@import gov.noaa.pfel.erddap.jte.Status
@import gov.noaa.pfel.erddap.jte.YouAreHere

@param YouAreHere youAreHere

@template.youarehere(youAreHere)

<pre>
Current time is ${Status.getCurrentTimeLocal()}
Startup was at ${Status.getStartupTimeLocal()}
Last major LoadDatasets started ${Status.getTimeSinceLastMajorLoadDatasetsStarted()} ago and@if(Status.isLoadDatasetsRunning()) is still running.@else finshed after ${Status.getLoadDatasetsElapsedTimeSeconds()} seconds.@endif
</pre>
```

`youarehere.jte`
```
@import gov.noaa.pfel.erddap.jte.YouAreHere

@param YouAreHere youAreHere

<h1 class="nowrap">
    <a title="${youAreHere.getErddapLinkTitle()}" rel="start"
        href="${youAreHere.getErddapUrl()}/index.html">
        ${youAreHere.getProgramName()}</a>
        &gt; ${youAreHere.getProtocol()} 
</h1>
```

Helper method in `EDStatic` to create a `YouAreHere` instance:

```
  /**
   * Create a YouAreHere model for JTE rendering.
   * 
   * @param request the request
   * @param language the index of the selected language
   * @param loggedInAs the logged in user
   * @param protocol the protocol (e.g., tabledap, griddap, etc.)
   * @return a YouAreHere model for JTE rendering
   */
  public static YouAreHere getYouAreHere(HttpServletRequest request, int language, String loggedInAs, String protocol) {
    return new YouAreHere(
        erddapUrl(request, loggedInAs, language),
        EDStatic.messages.clickERDDAPAr[language],
        protocol);
  }
```

And the actual status template rendering in `Erddap.java`:

```
      if (EDStatic.config.useHtmlTemplates) {
        // Use JTE template engine to render the status page
        TemplateEngine engine = TemplateEngine.createPrecompiled(ContentType.Html);
        YouAreHere youAreHere = EDStatic.getYouAreHere(request, language, loggedInAs, EDStatic.messages.statusAr[language]);
engine.render("status.jte", Map.of("youAreHere", youAreHere), new WriterOutput(writer));
      } else {
         ...
      }
``` 

This is just a start but I think it shows the intent:

* We want to do all HTML building in the JTE templates and not in code
* JTE model classes should be created which provide template access to data via static or instance methods

One other note, technically it's possible to access static methods on classes like `EDStatic` and `Calendar2` directly from the jte templates, but in my opinion using JTE model classes for all data access from templates is a good practice because it will make refactoring issues much easier to detect (they will fail/warn at compile time).

srstsavage: image

Forgot the screenshot to prove I'm not making all this up :joy:

WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
```diff
@@ -385,6 +385,10 @@ public class EDStatic {
       new ArrayList<>(); // keep here in case EmailThread needs to be restarted
   private static EmailThread emailThread;
 
+  public static EmailThread getEmailThread() {
```
ChrisJohnNOAA: Here and for the other get_Thread calls, I don't believe you need the thread object. Unless I missed something you only access their elapsed time. I'd recommend making a call to get current task elapsed time for each tread instead of a getter for them.

The reason is to control what can interact with the threads. I don't want random parts of code to be able to get the thread and change them.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
```diff
@@ -0,0 +1,292 @@
...
+import gov.noaa.pfel.erddap.util.*;
+import gov.noaa.pfel.erddap.util.EDStatic;
+import java.lang.management.ManagementFactory;
+
+public class Status {
+  private static Erddap erddap;
+
+  static {
+    try {
+      erddap = new Erddap();
```
ChrisJohnNOAA: You don't want to initialize a new Erddap, you'll want to pass in the existing one to this.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -4965,28 +4968,17 @@ public void doStatus(
...
-            out);
+        getHtmlWriterUtf8(request, language, loggedInAs, "status.html", queryString, "Status", out);
+
     try {
-      writer.write(
-          "<div class=\"standard_width\">\n"
-              + EDStatic.youAreHere(
-                  request, language, loggedInAs, EDStatic.messages.statusAr[language])
-              + "<pre>");
+      String youAreHereHtml =
```
ChrisJohnNOAA: It looks like this is only used if useHtmlTemplates is false. It should be created in that if statement.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -4997,24 +4989,29 @@ public void doStatus(
...
               + " dangerousMemoryEmails since last major LoadDatasets\n");
-      sb.append(Math2.memoryString() + " " + Math2.xmxMemoryString() + "\n\n");
-
+      sb.append(Math2.memoryString()).append(" ").append(Math2.xmxMemoryString()).append("\n\n");
       EDStatic.addCommonStatistics(sb);
       sb.append(traces);
-      writer.write(XML.encodeAsHTML(sb.toString()));
-      writer.write("</pre>\n");
+
+      String statisticsHtml = XML.encodeAsHTML(sb.toString());
```
ChrisJohnNOAA: It looks like this (and the above several lines of code) is only used if useHtmlTemplates is false. It should be created in that if statement.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -4997,24 +4989,29 @@ public void doStatus(
...
+
+      writer.write("<div class=\"standard_width\">\n");
```
ChrisJohnNOAA: Ideally any html will be in the templates not the Java code. Can this be moved so it's only used when useHtmlTemplates is false?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  public static long getLoadDatasetsElapsedTimeSeconds() {
+    return (EDStatic.lastMajorLoadDatasetsStopTimeMillis
+            - EDStatic.lastMajorLoadDatasetsStartTimeMillis)
+        / 1000;
+  }
+
+  public static boolean isLoadDatasetsRunning() {
+    // load datasets stop time > start time means the current load hasn't finished
```
ChrisJohnNOAA: I think the comment should read: load datasets stop time < start time means the current load hasn't finished

note the change from > to <

src/main/jte/status.jte
```diff
@@ -0,0 +1,49 @@
+@import gov.noaa.pfel.erddap.jte.Status
+<pre>
```
ChrisJohnNOAA, line 6: There should be a call to the youAreHere template from this.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -4997,24 +4989,29 @@ public void doStatus(
...
+
+      if (EDStatic.config.useHtmlTemplates) {
+        TemplateEngine engine = TemplateEngine.createPrecompiled(ContentType.Html);
+        engine.render("status.jte", Map.of(), new WriterOutput(writer));
```
ChrisJohnNOAA: Instead of Map.of(), should this be an instance of the Status class?

ChrisJohnNOAA (changes requested):
There's a lot of great progress. The new page is taking shape!

src/main/jte/status.jte
Harish-2003: Yes, actually i presently concentrated on building the status page will add it soon in the next commits

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
srstsavage: I need to review the current state of things tonight, but I believe all methods used by the template on `Status` are static so far, so there's no need for an instance of it.

ChrisJohnNOAA: I believe the status class need an instance of erddap provided to it in the constructor and so would need to be provided as a param to the template render call.

srstsavage: Ah ok, I see. It looks like the only information needed from the `Erddap` instance by Status are the sizes of the grid and table dataset maps. So, instead of passing the entire instance of `Erddap` and its 25k lines of complexity, let's just pass those map sizes to the `Status` constructor and store `int tableDatasetCount` and `int gridDatasetCount` instance variables on `Status`. We should still access all of the other status information from `EDStatic` statically. But you're right, after this change we would need to initialize an instance of `Status` and pass it to `engine.render`.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
srstsavage: Yep! This one's my fault :grin:

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
```diff
@@ -0,0 +1,292 @@
...
+
+  public static int getTaskListSize() {
+    return EDStatic.taskList.size();
+  }
+
+  public static String getTaskThreadStatusMessage() {
+    TaskThread thread = EDStatic.getTaskThread(); // use the getter
+    long tElapsedTime = thread == null ? -1 : thread.elapsedTime();
+    return tElapsedTime < 0
+        ? "Currently, no task is running.\n"
```
srstsavage: We don't want to be building any text messages in Java code, so that we can change the wording/formatting just by editing the template file. So instead of `getTaskThreadStatusMessage`, we'd want `getTaskThreadElapsedTimeMillis` returning the milliseconds as a `long` for if/then logic in the template and then `getTaskThreadElapsedTimeString` returning the `Calendar2` formatted string for display. Same for the other threads (email and touch) and the `getGcAndRequestShedStatus` method.

srstsavage (changes requested):

Harish-2003:
will try to solve/resolve the issues that are disscused in the comments in the next commit as of now i added the complete status page info (might have some mistakes) but will be trying to solve them asap!

Harish-2003:
image
image

WEB-INF/classes/gov/noaa/pfel/erddap/util/EDStatic.java
Harish-2003:
```  private static EmailThread emailThread;

  public static long getEmailThread() {
    return emailThread == null ? -1 : emailThread.elapsedTime();
  }
```
is this fine?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
Harish-2003: do you want me to pass erddap object in the status file ? like a param?

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
Harish-2003: correct me if i am wrong ! so you are suggesting to pass the values as params that are non static in the erddap rather then passing the whole object?
and then assiging those values to the variables in the constructor?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
Harish-2003: status.java file
```
  public static String getTaskThreadStatusMessage() {
    long tElapsedTime = EDStatic.getTaskThreadElapsedTime();
    return tElapsedTime < 0
        ? null
        : Calendar2.elapsedTimeString(tElapsedTime);
   }
```
status.jte
```
@if(Status.getTaskThreadStatusMessage()==null)Currently, no task is running.@elseThe current task has been running for ${Status.getTaskThreadStatusMessage()}@endif
```
edstatic.java 
```
  private static TaskThread taskThread;

  public long TaskThread getTaskThreadElapsedTime() {
    return taskThread == null ? -1 : taskThread.elapsedTime();
  }
```
i hope this works? if this is fine will try to make further changes for the remaining methods

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
ChrisJohnNOAA: As Shane pointed out, in the comment in Erddap.java, there are only a couple things the status page needs from the Erddap object. It's probably best just to pass in the grid datasets count and table datasets count rather than the entire Erddap object.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
ChrisJohnNOAA: I believe the recommendation is to pass the tableDatasetCount and gridDatasetCount as parameters to the Status constructor (and store those values in the status). Then pass the status object to the render call (where Map.of() currently is) as the parameter for the template.

I think this is what you were saying, but because there are multiple kinds of parameters involved I'm trying to be extra clear.

ChrisJohnNOAA (changes requested):

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
srstsavage: Looks right to me!

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
srstsavage: > pass the tableDatasetCount and gridDatasetCount as parameters to the Status constructor (and store those values in the status). Then pass the status object to the render call (where Map.of() currently is) as the parameter for the template.

Yep, exactly.

```
public class Status {
  private int gridDatasetCount;
  private int tableDatasetCount;
  
  public Status(int gridDatasetCount, int tableDatasetCount) {
    this.gridDatasetCount = gridDatasetCount;
    this.tableDatasetCount = tableDatasetCount;
  }

  public int getGridDatasetCount() {
    return gridDatasetCount;
  }

  public int getTableDatasetCount() {
    return tableDatasetCount;
  }
  ...
}
```

Erddap.java

```
      if (useHtmlTemplates(request)) {
        Status status = new Status(
            gridDatasetHashMap == null ? 0 : gridDatasetHashMap.size(),
            tableDatasetHashMap == null ? 0 : tableDatasetHashMap.size());
        TemplateEngine engine = TemplateEngine.createPrecompiled(ContentType.Html);
        engine.render("status.jte", Map.of("status", status), new WriterOutput(writer));
      } else {
        writer.write(youAreHereHtml + "<pre>" + statisticsHtml + "</pre>\n");
      }
```

status.jte

```
@import gov.noaa.pfel.erddap.jte.Status
@param Status status
<pre>
...
nGridDatasets = ${status.getGridDatasetCount()}
nTableDatasets = ${status.getTableDatasetCount()}
nTotalDatasets = ${status.getGridDatasetCount()+status.getTableDatasetCount()}
${Status.getdatasetsfailedtoload()}
@if(Status.getShowLoadErrorsOnStatusPage())${Status.getdatasetsfailedtoload()}@endif
```

Note that we still access the static Status methods in the template using `Status` but access the instance methods on the passed status param using `status`.

Harish-2003:
when  jte is false
image
when jte is true
image

Harish-2003:
i think i have made most of the changes that were suggested for the status page if i missed anything please let me know

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Status.java
```diff
@@ -0,0 +1,418 @@
...
+    return EDStatic.cldDatasetID;
+  }
+
+  public static String getDatasetLoadingTime() {
+    return Calendar2.elapsedTimeString(
+        Math2.longToDoubleNaN(System.currentTimeMillis() - EDStatic.cldStartMillis));
+  }
+
+  /*
+  public static ConcurrentHashMap<String, EDDGrid> getDatasetHashMap() {
```
ChrisJohnNOAA: Please remove the commented out code.

Harish-2003: done

srstsavage (approved):
In my view is this ready to merge now! :tada:

ChrisJohnNOAA (approved):

PR #309, opened 2025-06-16
Prometheus configs and grafana dashboards for monitoring.

ChrisJohnNOAA:
# Description

I've taken a pass at creating dashboards that may be useful for admins as well as a "fleet dashboard" that would be useful for seeing the state of the fleet of ERDDAP servers at a glance to aid with development decisions.

## Type of change

- [ ] I have made corresponding changes to the documentation (I'm working on a prometheus page for the documentation)

### DIFF
--- a/docker/prometheus/prometheus.yml
+++ b/docker/prometheus/prometheus.yml
@@ -19,17 +19,7 @@ scrape_configs:
     scheme: 'http'
     honor_labels: true
     static_configs:
-      - targets: ['erddap-erddap-1:8080']
+      - targets: ['host.docker.internal:8080']
         labels:
           instance: 'erddap-primary'
           region: 'primary'
-
-    # Example of monitoring multiple ERDDAP instances
-    # - job_name: 'erddap-secondary'
-    #   scrape_interval: 30s
-    #   metrics_path: '/erddap/metrics'
-    #   static_configs:
-    #     - targets: ['erddap-secondary:8080']
-    #       labels:
-    #         instance: 'erddap-secondary'
-    #         region: 'secondary'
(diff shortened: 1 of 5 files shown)


docker/prometheus/prometheus.yml
```diff
@@ -2,34 +2,24 @@ global:
...
-    # - job_name: 'erddap-secondary'
-    #   scrape_interval: 30s
-    #   metrics_path: '/erddap/metrics'
-    #   static_configs:
-    #     - targets: ['erddap-secondary:8080']
-    #       labels:
-    #         instance: 'erddap-secondary'
-    #         region: 'secondary'
+    static_configs:
+      - targets: ['host.docker.internal:8080']
```
srstsavage:
```suggestion
      - targets: ['erddap:8080']
```

The target should reference the erddap container directly by its service name, and not go through host routing.

srstsavage (changes requested):
This looks great, other than the prom target for the local erddap needing to change to `erddap:8080` (host port of the ERDDAP is configurable so we always want to scrape metrics using the stack networking and direct container hostname, also `host.docker.internal` doesn't work on Linux by default).

Eventually we should clean up a few things about the prometheus and grafana containers in the compose stack (not specify `container_name` and make the host port configurable with an env var like the ERDDAP port), but those issues weren't introduced here and can be handled in a separate PR.

srstsavage:
It does seem like the fleet metrics and dashboard might eventually fit better on a separate admin repo, but it's probably fine to include here for now.

docker/prometheus/prometheus.yml
ChrisJohnNOAA: Done

srstsavage (approved):

PR #312, opened 2025-06-29
added the jte version of legal.html page

Harish-2003:
# Addition of JTE Templates

- added the legal.html page in jte

working #295  

## Type of change

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
@@ -1758,18 +1758,32 @@ public class Erddap extends HttpServlet {
             "Legal Notices",
             out);
     try {
-      writer.write(
-          "<div class=\"standard_width\">\n"
-              + EDStatic.youAreHere(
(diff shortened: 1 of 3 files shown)


srstsavage:
This was my response to the above questions in slack:

> I think using EDStatic.messages.legalNoticesAr[language] as you have done in Legal.getLegalNoticesAr and wrapping in unsafe is good.
I see that EDStatic.messages.legal is a template in itself defined in messages.xml, with tokens like [standardContact] which get replaced by other messages values.

```
<legal><![CDATA[
[standardDisclaimerOfEndorsement]
[standardDisclaimerOfExternalLinks]
[standardPrivacyPolicy]
[standardDataLicenses]
[standardContact]
]]></legal>
```

> Because of this, I think we have to use EDStatic.legal as-is, like you've done in Legal.getLegal . So, I agree with your choices so far!

> @srstsavage should we file an issue where we collect these messages to come back to later and make it easier for admins make changes?

We can...this `legal` one seems to be a pretty special case, where the default `legal` template is basically just a list of other configuration parameters with no other content. So yeah, I agree that maintaining a list of areas to revisit after we fully support templates would be a good idea, either in an issue or elsewhere (GH project?).

PR #314, opened 2025-07-02
Mqtt client

ayushsingh01042003:
# Description

#### This will allow ERDDAP to build datasets in realtime by connecting to Mqtt brokers.

The following were added:
- EDDTableFromMqtt
- EDDTableFromMqttTests
- Parsing for the new dataset type EDDTableFromMqtt exlusive to the SAX parser
- HiveMQ client libraries
- GenerateDatasetsXml for EDDTableFromMqtt dataset


WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
```diff
@@ -0,0 +1,664 @@
...
+      boolean mustGetData)
+      throws Throwable {
+
+    if (!mustGetData)
+      // Just return a table with columns but no rows. There is never any metadata.
+      return Table.makeEmptyTable(sourceDataNames.toArray(), sourceDataTypes);
+
+    // read the file
+    Table table = new Table();
+    table.readJsonlCSV(tFileDir + tFileName, sourceDataNames, sourceDataTypes, false);
```
ChrisJohnNOAA: I know the EDDTableFromHttpGet uses file locks to avoid race conditions. Is there a reason you don't believe that's necessary here?

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
```diff
@@ -0,0 +1,664 @@
...
+        .sessionExpiryInterval(sessionExpiryInterval)
+        .simpleAuth()
+        .username(username)
+        .password(password != null ? password.getBytes() : null)
+        .applySimpleAuth()
+        .send()
+        .orTimeout(connectionTimeout, TimeUnit.SECONDS)
+        .thenApply(
+            connAck -> {
+              System.out.println("ERDDAP MQTT Client connected successfully!");
```
ChrisJohnNOAA: Either remove the System.out.println or change them to use String2.log (here and throughout the file).

src/test/java/gov/noaa/pfel/erddap/dataset/EDDTableFromMqttTests.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @TempDir private static Path TEMP_DIR;
+
+  @BeforeAll
+  static void init() {
+    Initialization.edStatic();
+  }
+
+  @org.junit.jupiter.api.Test
+  void testProcessMqttData() throws Throwable {
```
ChrisJohnNOAA, line 26: It would be good to cover some error cases in the tests as well. For example a message with different columns than what is expected to verify the behavior in those circumstances.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
```diff
@@ -0,0 +1,664 @@
...
+
+  /**
+   * Updates the in-memory fileTable with min/max statistics from a new batch of data. This method
+   * is ported from EDDTableFromHttpGet.
+   *
+   * @param newData a Table holding the new rows that were just written
+   * @param fullFileName the full path of the file that was written to
+   * @throws TimeoutException
+   */
+  public void updateFileTableWithStats(Table newData, String fullFileName) throws TimeoutException {
```
ChrisJohnNOAA: This is a long complex function. I think it's identical or nearly identical to the code in EDDTableFromHttpGet. Can this be a function on EDDTableFromFiles (since that owns fileTable) that both EDDTableFromMqtt and EDDTableFromHttpGet call?

ChrisJohnNOAA (changes requested):
Exciting to see MQTT coming together! I have a few comments.

kevin-obrien:
I am excited to see this come together as well, and I apologize for a late comment.  I just want to make that the plan is not only for ERDDAP so be able to ingest data, but also to notify data brokers that new data has been added to an ERDDAP.   This is important for real time data exchange under the new WIS 2 ecosystem

ChrisJohnNOAA:
> I am excited to see this come together as well, and I apologize for a late comment. I just want to make that the plan is not only for ERDDAP so be able to ingest data, but also to notify data brokers that new data has been added to an ERDDAP. This is important for real time data exchange under the new WIS 2 ecosystem

A dataset type based on MQTT ingested data is step 1. Notifying about data changes is complicated and currently something on the roadmap, but out of scope for this pull request.

@kevin-obrien when you say notify that new data has been added, do you mean literally just an MQTT message that indicates a change in the dataset (similar to the way the current subscribe system works), or do you mean a series of MQTT messages that detail everything that changed in the dataset?

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
ayushsingh01042003: Had referred a different class for this hence missed the lock, have added them since we would need it in this scenario.

src/test/java/gov/noaa/pfel/erddap/dataset/EDDTableFromMqttTests.java
ayushsingh01042003: Added 2 failure tests

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
```diff
@@ -652,13 +652,21 @@ public Table lowGetSourceDataFromFile(
...
+    fullFileName = String2.canonical(fullFileName);
+    ReentrantLock lock = String2.canonicalLock(fullFileName);
+    if (!lock.tryLock(String2.longTimeoutSeconds, TimeUnit.SECONDS))
+      throw new TimeoutException(
+          "Timeout waiting for lock on file in EDDTableFromMqtt.lowGetSourceDataFromFile: "
+              + fullFileName);
+    try {
+      Table table = new Table();
+      table.readJsonlCSV(fullFileName, sourceDataNames, sourceDataTypes, false);
+      table.standardize(standardizeWhat);
```
ChrisJohnNOAA: It's a best practice to hold the lock for as short as possible. In this case, there's no need for the standardize call to be within the lock. You can move the new table, standardize, and return out of the try and out of the lock.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
ChrisJohnNOAA: Do you have any questions on this?

ayushsingh01042003: I am trying to make this into a common function but there are subtle changes and also insertOrDelete being static and containing other code as well is making it a bit tricky.
While I do agree there is a quite a bit of reusable code here, is it good to make it common to the EDDTableFromFiles class in this scenario, seeing its just these two classes and also their subtle differences?

ChrisJohnNOAA: Are there any parts that can be shared? The calculate statistics section looks like a good candidate for that.

I'll need to look at the other sections more closely to see if there is more that's shareable.

ayushsingh01042003:
Made many assumptions for arguments here for EDDTableFromMqtt(including putting a random value for rows arg, since couldnt figure out what to place there).
Please let me know what changes are needed.

ChrisJohnNOAA:
> Made many assumptions for arguments here for EDDTableFromMqtt(including putting a random value for rows arg, since couldnt figure out what to place there). Please let me know what changes are needed.

I made a small change to simplify the params needed for the new method.

I also updated a couple of the parameters that MQTT passes down:
table.getColumns(), for the columnValues (instead of new PrimitiveArray[]).
table.nRows(), for the row count.

Please take a look and let me know what you think.

ayushsingh01042003:
The changes look good, simplified the arguments part pretty well. Any reason as to why we call statically using EDDTableFromFiles instead of the current Class name?

PR #320, opened 2025-07-17
Fix accumulation for aggregate rows where knowsActualRange is false

### LINKED ISSUE #319: Aggregate rows where children classes return false for `knowsActualRange()` don't accumulate ranges even when `actual_range` attributes are provided
**Describe the bug**
As in the title, if a class returns false for `knowsActualRange()`, it doesn't accumulate actual ranges. That would be understandable if the `actual_range` attribute wasn't set for my datasets

**To Reproduce**
1. Create a `datasets.xml` with an `EDDTableAggregateRows` with at least 2 children. Make the children that of EDD implementing class where `knowsActualRange()` is false (for example `EDDTableFromDatabase`) and provide an `actual_range` attribute for any variable. I've attached an example, but your own database values would need to be provided: [datasets.xml.zip](https://github.com/user-attachments/files/21274869/datasets.xml.zip) (zipped because xml is not allowed to be uploaded).
2. Navigate to the dataset and check the min and max ranges for the variables where `actual_range` was provided

**Expected behavior**
The ranges based on the actual_ranges in the attributes in `datasets.xml` would be the cumulative min and max:
  - time (translated by me to human readable format):
    - example_1: `2020-08-14T18:56:13Z` `2020-08-14T18:56:26Z`
    - example_2: `2022-08-24T19:21:33Z` `2022-08-24T19:23:20Z`
    - expected cumulative: `2020-08-14T18:56:13Z` `2022-08-24T19:23:20Z`
 - latitude:
    - example_1: `<att name="actual_range" type="doubleList">48.4251991 48.4252002</att>`
    - example_2: `<att name="actual_range" type="doubleList">48.3366411 48.3366528</att>`
    - expected cumulative: `48.3366411 48.4252002`
    - actual cumulative: `48.4251991 48.4252002`
 - longitude:
    - example_1: `<att name="actual_range" type="doubleList">-123.3718741 -123.371868</att>`
    - example_2: `<att name="actual_range" type="doubleList">-123.7499388 -123.7495656</att>`
    - expected cumulative: `-123.7499388 -123.371868`
    - actual cumulative: `-123.3718741 -123.371868`

**Screenshots**
Image

**Server**
Using gold standard. ERDDAP version 2.27. Tomcat 10.1.43. Java 21.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Any

**Additional context**
This was fixed for files and other dataset types [in this commit](https://github.com/ERDDAP/erddap/commit/d17fd4780139ff7f8de22d75350977cca017d169) in version 2.24

ItaloBorrelli:
# Description

Instead of checking `knowsActualRange`, use values of `destinationMin` and `destinationMax`.

Fixes #319 

## Type of change

- [ ] I have performed a self-review of my code


ItaloBorrelli:
I was considering removing the private variable `knowsActualRange` attribute, but it's used for the public implementation method `knowsActualRange()` for this class.

ItaloBorrelli:
> I believe one of the reasons for only trying to aggregate if all children have it is to avoid trying to aggregate a null or NaN value.

Understood. I'm trying to think of how to do this correctly then, though I'm having trouble explaining my thoughts so I'll write pseudocode.

### Current logic

As we know, it only sets `actual_range` if the `knowsActualRange` is true for all children. Misses all scenarios where variables have set `actual_range` attributes.
```
knowsActualRange = true

for each child dataset
  if !child.knowsActualRange then knowsActualRange = false

if knowsActualRange then
  for each variable
    for each child
      tMin = min(tMin, child.destinationMin())
      tMax = max(tMax, child.destinationMax())
  if !tMin.isMissingValue() && !tMax.isMissingValue() then
    set actual_range for variable (tMin, tMax)
```

### My code change

 - Doesn't do anything about `knowsActualRange`.
 - Doesn't check to see if all children have the variable range.
 - Min/Max with NaN will result in a Min/Max value being set.
```
knowsActualRange = true

for each child dataset
  if !child.knowsActualRange then knowsActualRange = false

for each variable
  for each child
    tMin = min(tMin, child.destinationMin())
    tMax = max(tMax, child.destinationMax())
  if !tMin.isMissingValue() && !tMax.isMissingValue() then
    set actual_range for variable (tMin, tMax)
```

### Possible update

I don't think that there should be a requirement that all children have `actual_range` for all variables, so I think it should be checked on a variable-by-variable basis. I know in our case we don't provide it. That might be what you were proposing.

I have a lot of differences here so before I commit a bunch of code I just want to check if you agree with my logic:
```
knowsActualRange = true

for each child dataset
  if !child.knowsActualRange then knowsActualRange = false

knowsRangeForAllChildrenAndVariables = true
for each variable
  knowsMinForVariable = true
  knowsMaxForVariable = true
  for each child
    cMin = child.destinationMin()
    cMax = child.destinationMax()
   
    // See note
    if cMin.isMissingValue() && !knowsActualRange then knowsMinForVariable = false
    if cMax.isMissingValue() && !knowsActualRange then knowsMaxForVariable = false
    tMin = min(tMin, cMin)
    tMax = max(tMax, cMax)

  // if any are not provided, use NaN
  tMin = knowsMinForVariable ? tMin : NaN
  tMax = knowsMaxForVariable ? tMax : NaN
  set actual_range for variable (tMin, tMax)

  if !knowsMinForVariable || !knowsMaxForVariable then knowsRangeForAllChildrenAndVariables = false

knowsActualRange = knowsActualRange || knowsRangeForAllChildrenAndVariables
```

My questions about this are:
 - Does `knowsActualRange` need to be updated? Or should it be based only on the original value?
 - Is it ok to set actual range where only one of min or max are known for all? Or do both need to be known?

Note: `&& !knowsActualRange` would be to ensure the change is not breaking, but I don't know if it's necessary. My concern is that if, say, all children are `EDDTableFromFile` but for some reason a variable has a missing value, if I were to just do `if cMin.isMissingValue() then knowsMinForVariable = false` then it could inadvertently impact someones expected `actual_range` for the aggregate rows.

ItaloBorrelli:
Thanks for the input! That helps a lot!

> One additional thought- Do we want to set min/max if some but not all of the children have values?
>
> I could see a benefit to the best possible value being provided.

Great question. I wasn't sure what would make the most sense. If we have 4 child datasets and their min max values for a particular variable are:

### Aggregate 1
```
<100,150>
<120,160>
<175,NaN>
```

### Aggregate 2
```
<100,150>
<120,160>
<NaN,NaN>
```

### Aggregate 3
```
<100,150>
<120,160>
<175,NaN>
<NaN,NaN>
```

|#|Description|Aggregate 1|Aggregate 2|Aggregate 3|Considerations|
|-|-|-|-|-|-|
|Option 1|If there are any children have min or max values that are `NaN`, accumulate to `NaN`.|`<100,NaN>`|`<NaN,NaN>`|`<NaN,NaN>`|Seems safe. There is no way that there are min values lower than the smallest provided in aggregate 1. In the other two, we can't be certain, and we don't prescribe a min or max where there could be lower or higher values for the unknown child ranges.|
|Option 2|Set min and max where "some but not all of the children have values". In this scenario, we find the min of all minimums and the maximum of all maximums, only using discrete values and ignoring all `NaN`s.|`<100,160>`|`<100,160>`|`<100,160>`|Aggregate 2 is probably fine, but 1 and 3 are suspect. The max is set as 160 but we know there is a dataset with values greater than that maximum.|
|Option 3|Ignore all ranges that are `<NaN,NaN>`. Minimum iff every remaining child has a minimum. Set maximum iff every remaining child has a maximum.|`<100,NaN>`|`<100,160>`|`<100,NaN>`|More complex and nuanced, but ensures any children with one or the other don't contribute to a mistake as in option 2. This doesn't mean that the minimum range value provided is actually the minimum, because children with range `<NaN,NaN>` could have values lower than cumulative minimum.||

I think that option 1 is most correct but least precise. Option 3 toes the line between correctness and precision. Option 2 may in some scenarios be incorrect, but is precise given the child dataset ranges.

I'm open to whichever you think makes the most sense and defer to your judgment @ChrisJohnNOAA. I assumed all minimums would be needed to set a minimum, ditto for maximums. Whichever you suggest I will go ahead with and I'll probably move it from a draft in the next commit unless I have any other questions for you.

ChrisJohnNOAA:
@ItaloBorrelli I was hoping to start the test process for the next release soon. I was curious if you had an ETA for this?

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableAggregateRows.java
```diff
@@ -486,24 +500,20 @@ public boolean lowUpdate(int language, String msg, long startUpdateMillis) throw
     PAOne tMin[] = new PAOne[ndv];
     PAOne tMax[] = new PAOne[ndv];
     EDDTable tChild = getChild(language, 0);
+    List<EDDTable> tChildren =
+        IntStream.range(0, nChildren).mapToObj(c -> getChild(language, c)).toList();
```
ChrisJohnNOAA: Is there a reason you can't use the List.of(tChildren) that you used above?

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableAggregateRows.java
```diff
@@ -365,21 +363,19 @@ public EDDTableAggregateRows(
...
+          accumulateActualRange(List.of(tChildren), dv, childVar.destinationDataPAType());
+
+      PAOne tMin = actualRange.getLeft();
+      PAOne tMax = actualRange.getRight();
+
+      PrimitiveArray pa = PrimitiveArray.factory(childVar.destinationDataPAType(), 2, false);
+      pa.addPAOne(tMin);
+      pa.addPAOne(tMax);
+      tSourceAtts.set("actual_range", pa);
+      tSourceAtts.set("actual_range", pa);
```
ChrisJohnNOAA: I assume the second set("actual_range" is not needed?

ChrisJohnNOAA (changes requested):
Thanks for tackling this!

A couple small comments on the changes.

As for other things- yes, tests would be great to verify the behavior. Since as you pointed at the documentation doesn't mention working/not working in specific circumstances any changes there are optional. 

I believe knowsActualRange and the function could be removed entirely from ERDDAP with this change. I'm happy to do that in a followup PR.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableAggregateRows.java
ItaloBorrelli: That's true. Thanks. Will remove.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableAggregateRows.java
ItaloBorrelli: In `lowUpdate` I think the concern is concurrency issues. That was the impression I got from the method definition and [the use of `getChild()` rather than `tChildren[0]`](https://github.com/ERDDAP/erddap/pull/320/files/3f64b421a65e545472f4e137310cb7a500a7c1ca#diff-44504e43a0af7f0d2c01730f085cfaa6e6f19bba4d49abc1c30a510cdb5894beR502). However, I just realized that if that is the case that `nChildren` may be incorrect.

ItaloBorrelli: Is that a valid concern? Does the `lowUpdate` for aggregate rows be concerned with the number of children? If so, I think I have to check if any of the children are null when fetched with `getChild(language, c)` and maybe even check if there are children past that point.

Or should I just use `tChildren` and assume it's fine for `lowUpdate`?

ChrisJohnNOAA: I believe the children array is not modified outside of the constructor so there should be no concurrency concerns around that.

I think it's fine to use `tChildren `for the `lowUpdate`.

ChrisJohnNOAA:
I think we can merge this as is. It'd be great to have db based tests using something like dbunit or other additional testing of the aggregate rows logic, but I understand integrating dbunit (or other db testing code) is a lot of work and and so I'd like to merge this and open an issue for the additional testing (unless you've got tests that will be ready soon).

PR #322, opened 2025-07-18
Mqtt client

ayushsingh01042003:
- Sample datasets.xml for EDDTableFromMqtt
- Using StringSubstitutorReader to process env variables in datasets.xml

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
@@ -213,7 +213,8 @@ public class LoadDatasets extends Thread {
             ReaderInputStream.builder()
                 .setReader(
                     new StringSubstitutorReader(
-                        new InputStreamReader(inputStream), new StringSubstitutor()))
+                        new InputStreamReader(getInputStream(inputStream)),
+                        new StringSubstitutor()))
                 .get();
       } else {
         inputStream = getInputStream(inputStream);


srstsavage:
Ouch...that's a pretty significant performance difference. The other option that comes to mind (I think we discussed during the last call as well) would be to check the element content (and probably attributes?) in the sax parser (`BaseDatasetHandler.endElement` and , looking for `${[a-zA-Z0-9_]+}` and replacing with an environment variable's value if one was found. Right now content is accessed directly from the StringBuilder in `BaseDatasetHandler` (`String contentStr = content.toString().trim();`), maybe right after that we do a check for that `${` prefix or above regex and replace accordingly?

WEB-INF/classes/gov/noaa/pfel/erddap/LoadDatasets.java
```diff
@@ -209,11 +208,16 @@ public void run() {
...
-          new StringSubstitutorReader(
-              new InputStreamReader(getInputStream(inputStream)), new StringSubstitutor());
-
-      inputStream = new ReaderInputStream(ssr, StandardCharsets.UTF_8);
+      if (EDStatic.config.enableEnvParsing) {
+        inputStream =
+            ReaderInputStream.builder()
+                .setReader(
+                    new StringSubstitutorReader(
+                        new InputStreamReader(inputStream), new StringSubstitutor()))
```
ChrisJohnNOAA: getInputStream wraps the stream in a BufferedInputStream, I think it'd be good to still have a buffer. It also handles the case where inputStream is null (which I believe is the case within the ERDDAP server). So for both of those reasons I'd recommend continuing to use getInputStream here.

ayushsingh01042003:
Made the change, had missed it. I assume this is what you meant @ChrisJohnNOAA ?

PR #323, opened 2025-07-20
addition table template and outofdatasets and subscriptions page

Harish-2003:
#Addition of table template and outofdatedatasets in jte

## Type of change


Harish-2003:
image
able to acheive some part will try to acheive the other part on further commits will be working on that

src/main/jte/outofdatedatasets.jte
```diff
@@ -0,0 +1,23 @@
+@import gov.noaa.pfel.erddap.jte.OutOfDateDatasets
+@import gov.noaa.pfel.erddap.jte.YouAreHere
+@import gov.noaa.pfel.erddap.jte.TableOptions
+@import gov.noaa.pfel.coastwatch.pointdata.Table;
+@param YouAreHere youAreHere
+@param int language
+@param TableOptions tableOptions
+@param Table table
+<head>
+    <link href="https://coastwatch.pfeg.noaa.gov/erddap/images/erddap2.css" rel="stylesheet" type="text/css">
```
ChrisJohnNOAA: I suspect this isn't needed here. The css file should be getting included in the startheadhtml. Also if we are going to include it we should use a relative path to load it from the local server, not coastwatch. I understand this is what the code used to do, but we should update it as part of this change.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/OutOfDateDatasets.java
```diff
@@ -0,0 +1,46 @@
+package gov.noaa.pfel.erddap.jte;
+
+import com.cohort.util.Calendar2;
+import gov.noaa.pfel.erddap.util.EDStatic;
+
+public class OutOfDateDatasets {
+  public static String getAdvcoutOfDateAr(int language) {
+    return EDStatic.messages.advc_outOfDateAr[language];
```
ChrisJohnNOAA: @srstsavage I think with the way messages currently work, there are going to be a LOT of functions like this. I wonder if having messages build a hashmap stringKey -> localizedMessage[], so that we can have a general function like:
EDStatic.messages.getLocalizedString(language, messageKey);

would be better?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,194 @@
...
+    this.timeColumn = timeColumn;
+    this.needEncodingAsHtml = needEncodingAsHtml;
+    this.allowWrap = allowWrap;
+    this.table = table;
+    int nCols = table.nColumns();
+    this.fileAccessBaseUrl = new String[nCols];
+    this.fileAccessSuffix = new String[nCols];
+  }
+
+  public long getTime() {
```
ChrisJohnNOAA: I didn't see this used. Did I miss it? Will it be used later? Several other functions here seem unused currently (getNoData, getSomethingWritten, maybe others). If there's no need for them, I'd prefer not to add the methods.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/OutOfDateDatasets.java
srstsavage: Wow, yeah, looks like ~1000 message strings? I hate to even say it out loud but would it be overkill to use an enum for the hashmap key? If we're just feeding a freeform string messageKey to `getLocalizedString` I don't think we would catch typos until runtime right?

ChrisJohnNOAA: Yeah, an enum to keep the build time checks sounds good.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
Harish-2003: will try to look and if there is no use of them will remove them

WEB-INF/classes/gov/noaa/pfel/erddap/jte/OutOfDateDatasets.java
srstsavage: Been thinking about this a bit more. I'm wondering if we should just reference messages strings directly in the JTE templates, at least for the scope of this GSoC JTE project. So instead of making wrapper function in the JTE model classes (example `OutOfDateDatasets.getAdvcoutOfDateAr`, we just import EDStatic in the template and then reference the messages string directly:

```
@import gov.noaa.pfel.erddap.util.EDStatic
...
${EDStatic.messages.advc_outOfDateAr[language]}
```

For most template values we want to hide complexity behind a model class function, but messages are a bit of a special case since they're so direct and there are so many of them. Building the above mentioned hashmap and enum is going to introduce a lot of busy work and room for error, and I'm concerned that it would slow down progress. Any thoughts on just referencing the `EDStatic.messages` strings directly in the templates for now? It looks like `Legal` is the only place other than `OutOfDataDatasets` where we've already created messages wrapper functions, and there are only two.

ChrisJohnNOAA: > Been thinking about this a bit more. I'm wondering if we should just reference messages strings directly in the JTE templates, at least for the scope of this GSoC JTE project. So instead of making wrapper function in the JTE model classes (example `OutOfDateDatasets.getAdvcoutOfDateAr`, we just import EDStatic in the template and then reference the messages string directly:
> 
> ```
> @import gov.noaa.pfel.erddap.util.EDStatic
> ...
> ${EDStatic.messages.advc_outOfDateAr[language]}
> ```
> 
> For most template values we want to hide complexity behind a model class function, but messages are a bit of a special case since they're so direct and there are so many of them. Building the above mentioned hashmap and enum is going to introduce a lot of busy work and room for error, and I'm concerned that it would slow down progress. Any thoughts on just referencing the `EDStatic.messages` strings directly in the templates for now? It looks like `Legal` is the only place other than `OutOfDataDatasets` where we've already created messages wrapper functions, and there are only two.

That's a fine plan.

Though I did start on the migration to using enums for messages. I think it will be a nice improvement. We should not hold the JTE project up for it.

srstsavage: @Harish-2003 Just to recap, the plan is for JTE templates to access `EDStatic.messages` for now:

```
@import gov.noaa.pfel.erddap.util.EDStatic
...
${EDStatic.messages.advc_outOfDateAr[language]}
```

and Chris will update these references to use the new hashmap/enum method when he's ready to merge that work. Don't bother creating wrapper methods for `EDStatic.messages` in the JTE data model classes anymore. Thanks!

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,190 @@
...
+  public boolean checkLengthFile(int col) {
+    return (fileAccessBaseUrl[col].length() > 0 || fileAccessSuffix[col].length() > 0);
+  }
+
+  public String getTs(String s) {
+    return needEncodingAsHtml ? s : XML.decodeEntities(s);
+  }
+
+  public String getUrl(int col, String s) {
+    return fileAccessBaseUrl[col] + getTs(s) + fileAccessSuffix[col];
```
Harish-2003, line 175: here i used the direct syntax rather than using 
String url =
                  XML.encodeAsHTMLAttribute(fileAccessBaseUrl[col] + ts + fileAccessSuffix[col]);
                  do you want me to do the same for remaining or do you want me to leave as it is ???

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,190 @@
...
+  public List<String> getExtractUrls(String s) {
+    return String2.extractUrls(s);
+  }
+
+  public boolean getContainsUrl(String s) {
+    return String2.containsUrl(s);
+  }
+
+  public String encodeaddhttps(String s) {
+    return XML.encodeAsHTMLAttribute(String2.addHttpsForWWW(s));
```
Harish-2003, line 202: here !?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  public boolean checkDecodeEntities(String s) {
+    return !needEncodingAsHtml && String2.isUrl(XML.decodeEntities(s));
+  }
+
+  public String getEncodeashtmlattribute(String s) {
+    return needEncodingAsHtml ? XML.encodeAsHTMLAttribute(s) : s;
```
Harish-2003, line 210: here?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
srstsavage: Yes, I think its correct to use the direct approach and not use `XML.encodeAsHTMLAttribute` since JTE automatically escapes HTML attributes.

https://jte.gg/html-rendering/#html-escaping

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
srstsavage: Same as above, so this method could probably be changed to `addHttps` and just return `String2.addHttpsForWWW(s)`, or `String2.addHttpsForWWW` could be called directly.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
srstsavage: As above, I think this method is probably not needed at all since JTE should automatically handle HTML attribute escaping.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -14082,24 +14089,65 @@ public void doOutOfDateDatasets(
...
+        writer.write("<div class=\"standard_width\">");
+        writer.write(EDStatic.youAreHere(request, language, loggedInAs, shortTitle));
+        writer.write(XML.encodeAsHTML(EDStatic.messages.get(Message.ADVC_OUT_OF_DATE, language)));
+        writer.write("\n<p>");
+        if (table.nRows() == 0) {
+          writer.write(
+              "["
+                  + MessageFormat.format(EDStatic.messages.get(Message.N_MATCHING, language), "0")
+                  + " "
+                  + EDStatic.messages.get(Message.ADVC_OUT_OF_DATE, language)
```
ChrisJohnNOAA: I believe this is supposed to be ADVN_OUT_OF_DATE

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -14082,24 +14089,65 @@ public void doOutOfDateDatasets(
...
+              null,
+              1, // other classes, bgColor, border,
+              false,
+              mtCol,
+              false, // writeUnits, timeColumn, needEncodingAsHtml,
+              false); // allowWrap
+        }
+
+        // autoRefresh message
         writer.write(
```
ChrisJohnNOAA, line 14148: Is this supposed to be here? I think it's the same as the above in the else block? Was it duplicated in the merge?

Harish-2003: i think so !! i just made another commit

src/main/jte/outofdatedatasets.jte
```diff
@@ -0,0 +1,22 @@
...
+@param int language
+@param TableOptions tableOptions
+
+@template.youarehere(youAreHere)
+
+${OutOfDateDatasets.getMessage(Message.ADVC_OUT_OF_DATE,language)}
+
+@if(tableOptions.getNRows() == 0)
+    [${OutOfDateDatasets.getFormated(OutOfDateDatasets.getMessage(Message.N_MATCHING,language),"0")}<br>
+    ${OutOfDateDatasets.getMessage(Message.ADVC_OUT_OF_DATE,language)}]
```
srstsavage:
```suggestion
    ${OutOfDateDatasets.getMessage(Message.ADVN_OUT_OF_DATE,language)}]
```

I think this is also supposed to be `ADVN_OUT_OF_DATE` right?

src/main/jte/Table.jte
```diff
@@ -0,0 +1,72 @@
+@import gov.noaa.pfel.erddap.jte.TableOptions
```
srstsavage, line 1: Unless there is a reason not to, can we move `src/main/jte/Table.jte` to `src/main/jte/table.jte` for consistency?

src/main/jte/subscription.jte
```diff
@@ -0,0 +1,38 @@
+@import gov.noaa.pfel.erddap.jte.YouAreHere
+@import gov.noaa.pfel.erddap.util.EDStatic
+@import gov.noaa.pfel.erddap.jte.Subscription
+@import gov.noaa.pfel.erddap.util.EDMessages.Message;
+@param YouAreHere youAreHere
+@param int language
+@template.youarehere(youAreHere)
+${Subscription.getMessage(Message.SUBSCRIPTION_0_HTML,language)}
```
srstsavage: Everything after youarehere should be wrapped in a `<div class="standard_width">`

src/main/jte/outofdatedatasets.jte
```diff
@@ -0,0 +1,22 @@
+@import gov.noaa.pfel.erddap.jte.OutOfDateDatasets
+@import gov.noaa.pfel.erddap.jte.YouAreHere
+@import gov.noaa.pfel.erddap.jte.TableOptions
+@import gov.noaa.pfel.erddap.util.EDStatic
+@import gov.noaa.pfel.erddap.util.EDMessages.Message
+@param YouAreHere youAreHere
+@param int language
+@param TableOptions tableOptions
+
```
srstsavage, line 9: Everything on the page should be wrapped in `<div class="standard_width">`

src/main/jte/outofdatedatasets.jte
```diff
@@ -0,0 +1,22 @@
...
+@template.youarehere(youAreHere)
+
+${OutOfDateDatasets.getMessage(Message.ADVC_OUT_OF_DATE,language)}
+
```
srstsavage, line 14: The rest of the page should be wrapped in `<p>`

src/main/jte/outofdatedatasets.jte
```diff
@@ -0,0 +1,22 @@
...
+@if(tableOptions.getNRows() == 0)
+    [${OutOfDateDatasets.getFormated(OutOfDateDatasets.getMessage(Message.N_MATCHING,language),"0")}<br>
```
srstsavage: If the page is wrapped in `<p>` I believe the trailing `<br>` can be removed

WEB-INF/classes/gov/noaa/pfel/erddap/jte/OutOfDateDatasets.java
```diff
@@ -0,0 +1,20 @@
+package gov.noaa.pfel.erddap.jte;
+
+import com.cohort.util.Calendar2;
+import gov.noaa.pfel.erddap.util.EDMessages.Message;
+import gov.noaa.pfel.erddap.util.EDStatic;
+import java.text.MessageFormat;
+
+public class OutOfDateDatasets {
```
srstsavage: Since none of the methods in `OutOfDateDatasets` are specific to out-of-date datasets, I think it may be better to move all of these methods to a `TemplateUtil` or `TemplateHelper` class that can be reused in multiple templates.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Subscription.java
```diff
@@ -0,0 +1,32 @@
+package gov.noaa.pfel.erddap.jte;
+
+import gov.noaa.pfel.erddap.util.EDMessages.Message;
+import gov.noaa.pfel.erddap.util.EDStatic;
+import gov.noaa.pfel.erddap.util.Subscriptions;
+import java.text.MessageFormat;
+
+public class Subscription {
+  public static String getFormated(String value, String val) {
```
srstsavage: Consolidate method to `TemplateUtil` and access there.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Subscription.java
```diff
@@ -0,0 +1,32 @@
...
+    return MessageFormat.format(value, val);
+  }
+
+  public static String getMessage(Message message, int language) {
```
srstsavage: Consolidate method to `TemplateUtil` and access there.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/Subscription.java
```diff
@@ -0,0 +1,32 @@
...
+    return EDStatic.messages.get(message, language);
+  }
+
+  public static String getAddHtml() {
```
srstsavage: Not a big deal to leave like this, but I'm not sure why we don't just access `Subscriptions.VALIDATE_HTML` etc directly in the template. If we did that and moved the generic methods to `TemplateUtil`, this class could be removed entirely.

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,178 @@
...
+  private final boolean writeUnits;
+  private final int timeColumn;
+  private final boolean needEncodingAsHtml;
+  private final boolean allowWrap;
+  private final Table table;
+  private boolean somethingWritten = false;
+  private final String[] fileAccessBaseUrl;
+  private final String[] fileAccessSuffix;
+
+  public TableOptions(
```
srstsavage: This is optional, but it would be nice to have a `TableOptions.Builder` to make the creation of a new `TableOptions` clearer.

https://www.digitalocean.com/community/tutorials/builder-design-pattern-in-java

srstsavage:
When testing `/erddap/outOfDateDatasets.html`, I notice that I can't use the `jte` boolean query parameter to switch JTE on and off because `jte` is interpreted as a data constraint:

/erddap/outOfDateDatasets.html?jte=false

```
Error {
    code=400;
    message="Bad Request: Query error: All constraints (including \"jte=...\") must be preceded by '&'.";
}
```

I was wondering if it was possible to remove `jte` from the request parameters after it's detected, but that's probably undesirable in the rare case that someone had a `jte` data variable. We could also check for some HTTP header (e.g. `X-Use-JTE`) but that's going to be a pain to use in the browser/would require use of a browser plugin. Maybe that's better than nothing. I'll think on it, let me know if anyone has ideas.

src/main/jte/Table.jte
```diff
@@ -0,0 +1,72 @@
...
+@import gov.noaa.pfel.coastwatch.pointdata.Table
+@param TableOptions tableOptions
+@if(tableOptions.getNRows()==0)
+    <p>${tableOptions.getNoData()} (nRows =0)</p>
+@else
+<table class="erd${tableOptions.getWrap()}${tableOptions.getOtherClasses()}" style="background-color:${tableOptions.getBgColor()};">
+
+
```
srstsavage:
```suggestion

```

Whitespace cleanup 😁

src/main/jte/outofdatedatasets.jte
```diff
@@ -0,0 +1,25 @@
...
+@import gov.noaa.pfel.erddap.jte.YouAreHere
+@import gov.noaa.pfel.erddap.jte.TableOptions
+@import gov.noaa.pfel.erddap.util.EDStatic
+@import gov.noaa.pfel.erddap.util.EDMessages.Message
+@param YouAreHere youAreHere
+@param int language
+@param TableOptions tableOptions
+
+@template.youarehere(youAreHere)
+<div class ="standard_width">
```
srstsavage:
```suggestion
<div class="standard_width">
```

Small whitespace cleanup.

src/main/jte/subscription.jte
```diff
@@ -0,0 +1,41 @@
+@import gov.noaa.pfel.erddap.jte.YouAreHere
+@import gov.noaa.pfel.erddap.util.EDStatic
+@import gov.noaa.pfel.erddap.jte.TemplateHelper
+@import gov.noaa.pfel.erddap.util.EDMessages.Message
+@import gov.noaa.pfel.erddap.util.Subscriptions
+@param YouAreHere youAreHere
+@param int language
+@template.youarehere(youAreHere)
+<div class ="standard_width">
```
srstsavage, line 9:
```suggestion
<div class="standard_width">
```

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+import com.cohort.util.String2;
+import com.cohort.util.XML;
+import gov.noaa.pfel.coastwatch.pointdata.Table;
+import java.util.List;
+
+public class TableOptions {
+
+  private final String otherClasses;
+  private final String bgColor;
+  private final int border;
```
srstsavage, line 15: It looks like `border` is never used (there's no getter in `TableOptions` for it). `border` is also never used in `Table.saveAsHtml`, so we should probably remove it there as well! @ChrisJohnNOAA any objection?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+
+  public int getNColumns() {
+    return table.nColumns();
+  }
+
+  public int getNRows() {
+    return table.nRows();
+  }
+
+  public String insideForLoophead(int col) {
```
srstsavage, line 117:
```suggestion
  public String getColumnHeader(int col) {
```

Let's rename this method to `getColumnHeader`

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+    if (!String2.isSomething(fileAccessSuffix[col])) fileAccessSuffix[col] = "";
+
+    return s;
+  }
+
+  public boolean getWriteUnits() {
+    return writeUnits;
+  }
+
+  public String insideForLoopUnits(int col) {
```
srstsavage, line 141:
```suggestion
  public String getColumnUnits(int col) {
```

Let's rename this method to `getColumnUnits`

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+      tUnits = "&nbsp;";
+    }
+    return tUnits;
+  }
+
+  public int getTimeColumn() {
+    return timeColumn;
+  }
+
+  public String getDoubleDataConverted(int col, int row) {
```
srstsavage, line 158:
```suggestion
  public String getTimeString(int col, int row) {
```

Method rename for clarity

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+
+  public String getStringData(int col, int row) {
+    return table.getStringData(col, row);
+  }
+
+  public boolean checkLengthFile(int col) {
+    return (fileAccessBaseUrl[col].length() > 0 || fileAccessSuffix[col].length() > 0);
+  }
+
+  public String getTs(String s) {
```
srstsavage, line 170:
```suggestion
  private String getTs(String s) {
```

I believe this can be a private method?

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+    return needEncodingAsHtml ? s : XML.decodeEntities(s);
+  }
+
+  public String getUrl(int col, String s) {
+    return fileAccessBaseUrl[col] + getTs(s) + fileAccessSuffix[col];
+  }
+
+  public String getEncodeashtml(String s) {
```
srstsavage, line 178:
```suggestion
  public String getEncodeAsHtml(String s) {
```

Method name case fixing

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+    return needEncodingAsHtml ? XML.encodeAsHTML(s) : s;
+  }
+
+  public boolean checkcontainsurl(String s) {
```
srstsavage, line 182:
```suggestion
  public boolean checkContainsUrl(String s) {
```

Method name case fixing

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+    return needEncodingAsHtml && String2.containsUrl(s);
+  }
+
+  public boolean checkmouseover(String s) {
```
srstsavage, line 186:
```suggestion
  public boolean checkMouseOver(String s) {
```

Method name case fixing

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+
+  public List<String> getExtractUrls(String s) {
+    return String2.extractUrls(s);
+  }
+
+  public boolean getContainsUrl(String s) {
+    return String2.containsUrl(s);
+  }
+
+  public String encodeaddhttps(String s) {
```
srstsavage, line 201:
```suggestion
  public String encodeAddHttps(String s) {
```

Method name case fixing

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
```diff
@@ -0,0 +1,225 @@
...
+    return XML.encodeAsHTMLAttribute(String2.addHttpsForWWW(s));
+  }
+
+  public boolean checkDecodeEntities(String s) {
+    return !needEncodingAsHtml && String2.isUrl(XML.decodeEntities(s));
+  }
+
+  public String getEncodeashtmlattribute(String s) {
```
srstsavage, line 209:
```suggestion
  public String getEncodeAsHtmlAttribute(String s) {
```

Method name case fixing

src/main/jte/Table.jte
srstsavage: Looks like this rename still needs to happen (`outofdatedatasets.jte` references `@template.table` but the table template is still named `Table.jte`)

WEB-INF/classes/gov/noaa/pfel/erddap/jte/TableOptions.java
ChrisJohnNOAA: Yeah, if it's never used go ahead and remove it. Looking at the code, it looks like it's passed down through several layers, we can likely remove it through all of them.

PR #331, opened 2025-07-29
add github as a docker image provider

ocefpaf:
# Description

Some people are having trouble with dockerhub today and having an alternative source for the images could help.

Note that this PR is untested and I'm not familiar with the docker image publication here. Please feel free to close it if this is not planned or desired for this repo. 

Fixes # (issue)

## Type of change

Added a new docker image source.

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -75,6 +75,7 @@ jobs:
       - name: Build and push Docker image
         uses: docker/build-push-action@v5
         with:
+          platforms: linux/amd64,linux/arm64
           context: .
           file: ./Dockerfile
           push: true
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -61,6 +61,13 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PUSH_TOKEN }}
 
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: erddap
+          password: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Get current date for tag
         id: date
         run: echo "::set-output name=now::$(date +'%Y%m%d%H%M%S')"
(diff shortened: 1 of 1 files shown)


abkfenris:
Is there still some things that need to happen locally for a release build (files not in repo), or could we build a release and push on tagging a commit?

ocefpaf:
> I'm ok with adding this, but we should probably also make the versioned releases available there as well.

I'm happy to help or wait for that. This is the pattern I use for tag/release publication with GHA.

https://github.com/ioos/erddapy/blob/188bb9395e6f62e9ffe4033eda89d93f26ab72bb/.github/workflows/pypi.yml#L3-L10

and

https://github.com/ioos/erddapy/blob/188bb9395e6f62e9ffe4033eda89d93f26ab72bb/.github/workflows/pypi.yml#L61

Not sure if it helps and/or in the same direction of what you are planning.

ChrisJohnNOAA:
> Is there still some things that need to happen locally for a release build (files not in repo), or could we build a release and push on tagging a commit?

We could release and push on tagging a commit. Though we would also need to add support for arm builds (I haven't investigated this yet, so it might be easy).

abkfenris:
> > Is there still some things that need to happen locally for a release build (files not in repo), or could we build a release and push on tagging a commit?
> 
> We could release and push on tagging a commit. Though we would also need to add support for arm builds (I haven't investigated this yet, so it might be easy).

With QEMU already setup it should be adding another key to the platform action.

https://docs.docker.com/build/ci/github-actions/multi-platform/

```diff
 name: Build and push Docker image
        uses: docker/build-push-action@v5
        with:
+         platforms: linux/amd64,linux/arm64
          context: .
          file: ./Dockerfile
          push: true
          tags: |
            erddap/erddap:alpha-latest
            erddap/erddap:${{ steps.date.outputs.now }}
```

ocefpaf:
> We could release and push on tagging a commit. Though we would also need to add support for arm builds (I haven't investigated this yet, so it might be easy).

Those two lines I sent above will do that. +1 for [QEMU](https://github.com/ocefpaf/netcdf-manylinux/blob/4e470f1e0727e1a9451b0197b54624d8a7022b76/.github/workflows/build-image.yml#L44-L47), it is slow but works fine.

ChrisJohnNOAA (approved):

PR #333, opened 2025-07-30
Release Docker images

abkfenris:
# Description

This builds on #331 to enable building of Docker images on tagging 
Releases.

Using the [Docker Metadata Action](https://github.com/docker/metadata-action), it generates a different set of tags 
as needed based on the event (push commit to main, push tag to main).

When a commit is pushed to main, it should generate `alpha-latest` and 
date tagged images. When a (semver compatible) tag is pushed, it should 
generate an image that has that tag.

Fixes # https://github.com/ERDDAP/erddap/pull/331#issuecomment-3134211512

## Type of change

- [X] New feature (non-breaking change which adds functionality)


.github/workflows/on-push.yml
```diff
@@ -3,6 +3,8 @@ name: Push Build and test ERDDAP
 on:
   push:
     branches: [ "main" ]
+    tags:
+      - 'v*'
```
ocefpaf, line 7: I have a slightly preference for:

```
  release:
    types:
      - published
```

That would force the tag to always start with a `v` and surprise us if we ever drop that in the future.

I can see people wanting to publish the docker image in a tag tough, that way the docker image info can be part of the release notes.

abkfenris: The metadata action doesn't directly support release events, and looking through the issues it may be a little more unreliable what it generates compared to pushing a tagged commit.

This way supports both manually pushing a tag then making a release, and drafting a release and letting Github create the tag.

srstsavage:
https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/

ChrisJohnNOAA:
> Whoops, I thought I marked this ready to review earlier, say before Github releases something relevant.
> 
> > https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/
> 
> From what I remember, building across multiple jobs and getting them combined correctly on hubs is a bit tricky still. It looks like this is the process https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners but I don't have experience with that directly (and not a ton of time to tinker with it for now). Maybe we want to try that separately?

Yeah, I'm happy to merge this and tinker more later (whoever has time). I think the current arm build was acceptable, so I think doing this can wait.

PR #339, opened 2025-08-07
Zizmor lints

ocefpaf:
This PR implements all zizmor suggestion's and add a lint to check for future changes.

[Zizmor is a GitHub Actions security inspection](https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection). If you think this is too noisy to have it here please feel free to close the PR.

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -22,6 +22,6 @@ jobs:
           persist-credentials: false
 
       - name: Run zizmor 🌈
-        uses: zizmorcore/zizmor-action@f52a838cfabf134edcbaa7c8b3677dde20045018 # v0.1.1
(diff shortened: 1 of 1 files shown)


ChrisJohnNOAA:
I'm seeing multiple [Anonymous Definition](https://docs.zizmor.sh/audits/#anonymous-definition) in the report: https://github.com/ERDDAP/erddap/security/code-scanning?query=pr%3A339+is%3Aopen

It looks to me like we have name set appropriately. Is there something I missed? If not, I'd prefer to not have false positives, should we use the regular persona?

ocefpaf:
> https://github.com/ERDDAP/erddap/security/code-scanning?query=pr%3A339+tool%3Azizmor+is%3Aopen

I believe these results are not visible to non-members. I get a 404 for that URL.

ChrisJohnNOAA (approved):
Now there's no warnings.

PR #341, opened 2025-08-09
Mqtt broker

ayushsingh01042003:
MQTT broker integration for ERDDAP:

- ContextListener to invoke and close broker
- sample `config.xml` sets default configuration to the broker
- Check to allow empty directories
- Flags for broker directories

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/MqttBrokerContextListener.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/MqttBrokerContextListener.java
@@ -51,9 +51,10 @@ public class MqttBrokerContextListener implements ServletContextListener {
 
   @Override
   public void contextDestroyed(ServletContextEvent sce) {
-    if (enableMqttBroker && this.hiveMQ != null) {
+    if (this.hiveMQ != null) {
       try {
         this.hiveMQ.stop().join();
+        this.hiveMQ = null;
       } catch (final Exception ex) {
         System.out.println("Error stopping MQTT broker");
         ex.printStackTrace();
--- a/development/jetty/config/setup.xml
+++ b/development/jetty/config/setup.xml
@@ -388,8 +388,6 @@ If an attribute is a global attribute, identify it by prefixing it with "global:
 <categoryAttributes>global:cdm_data_type, global:institution, ioos_category, global:keywords, long_name, standard_name, variableName</categoryAttributes>
 
 <enableMqttBroker>true</enableMqttBroker>
-<mqttConfigFolder>WEB-INF/classes/gov/noaa/pfel/erddap/config</mqttConfigFolder>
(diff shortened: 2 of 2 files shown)


WEB-INF/classes/gov/noaa/pfel/erddap/MqttBrokerContextListener.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Override
+  public void contextInitialized(ServletContextEvent sce) {
+    if (enableMqttBroker) {
+      try {
+        EmbeddedHiveMQBuilder embeddedHiveMQBuilder = EmbeddedHiveMQ.builder();
+
+        if (EDStatic.config.mqttConfigFolder != null
+            && !EDStatic.config.mqttConfigFolder.isEmpty()) {
+          embeddedHiveMQBuilder =
```
ChrisJohnNOAA, line 25: Will Hive MQTT happily startup with default values or are any of these configs needed? I was wondering if we should require any of them (and log a message if enableMqttBroker == true and they are unset? If Hive has good defaults and works well without them that is not needed.

WEB-INF/classes/gov/noaa/pfel/erddap/MqttBrokerContextListener.java
```diff
@@ -0,0 +1,63 @@
...
+      } catch (final Exception ex) {
+        System.out.println("Error starting MQTT broker: ");
+        ex.printStackTrace();
+      }
+    }
+  }
+
+  @Override
+  public void contextDestroyed(ServletContextEvent sce) {
+    if (enableMqttBroker && this.hiveMQ != null) {
```
ChrisJohnNOAA: I'd recommend not checking if the broker is enabled, but just if it's instantiated (in case we started it up, but then that value got changed).

Also once we shut it down, set it to null.

```suggestion
if (this.hiveMQ != null) {
      try {
        this.hiveMQ.stop().join();
        this.hiveMQ = null;
      } catch (final Exception ex) {
        System.out.println("Error stopping MQTT broker");
        ex.printStackTrace();
      }
    }
```

ChrisJohnNOAA (changes requested):

WEB-INF/classes/gov/noaa/pfel/erddap/MqttBrokerContextListener.java
ayushsingh01042003: HiveMQ does start fine with default values. These checks here were implemented for that. Should i remove the `config.xml` then? Since it wont be needed

ChrisJohnNOAA: I think it will be good to have the sample xml file so it's easy for folks to grab it and modify it. I'd say leave it. I do need to think about where we want to have it though.

PR #345, opened 2025-08-16
Pubish dataset change to local broker

ayushsingh01042003:
- Flag to enable publishing change to broker
- tryToDoActions checks to publish to broker
- publishData method

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
@@ -24460,6 +24460,14 @@ widgets.select("frequencyOption", "", 1, frequencyOptions, frequencyOption, "") 
       try {
         StringArray actions = null;
 
+        // publish change to local broker, if enabled
+        if (EDStatic.config.publishMqttNotif) {
+          // check if client is configured and broker is localhost
+          if (EDDTableFromMqtt.asyncClient != null && EDStatic.config.enableMqttBroker) {
+            EDDTableFromMqtt.publishData(change, "change/" + tDatasetID);
+          }
+        }
+
         if (EDStatic.config.subscriptionSystemActive) {
           // get subscription actions
           try { // beware exceptions from subscriptions
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
@@ -38,6 +38,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 public class EDDTableFromMqtt extends EDDTableFromFiles {
+  public static Mqtt5AsyncClient asyncClient;
 
   public static final int DEFAULT_STANDARDIZEWHAT = 0;
   private static final int MQTT_PORT = 1883;
@@ -193,9 +194,9 @@ public class EDDTableFromMqtt extends EDDTableFromFiles {
             sessionExpiryInterval,
             connectionTimeout,
             automaticReconnect);
-    Mqtt5AsyncClient asyncClient = response.join();
+    asyncClient = response.join();
 
-    subscribeToDatasetTopics(asyncClient, topics, MqttQos.AT_LEAST_ONCE);
+    subscribeToDatasetTopics(topics, MqttQos.AT_LEAST_ONCE);
   }
 
   public static CompletableFuture<Mqtt5AsyncClient> initialiseMqttAsyncClient(
@@ -263,13 +264,17 @@ public class EDDTableFromMqtt extends EDDTableFromFiles {
             });
   }
 
-  public void subscribeToDatasetTopics(Mqtt5AsyncClient client, String[] topics, MqttQos qosLevel) {
+  public static void publishData(String data, String topic) {
+    asyncClient.publishWith().topic(topic).payload(data.getBytes()).send();
+  }
+
+  public void subscribeToDatasetTopics(String[] topics, MqttQos qosLevel) {
 
     for (String topic : topics) {
       if (topic == null || topic.trim().isEmpty()) {
         continue;
       }
-      client
+      asyncClient
           .subscribeWith()
           .topicFilter(topic)
           .qos(qosLevel)
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/util/EDConfig.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/util/EDConfig.java
@@ -241,6 +241,7 @@ public class EDConfig {
 
   @FeatureFlag public final boolean variablesMustHaveIoosCategory;
   @FeatureFlag public boolean useSaxParser;
+  @FeatureFlag public boolean publishMqttNotif;
   @FeatureFlag public boolean enableEnvParsing;
   @FeatureFlag public boolean updateSubsRssOnFileChanges;
   @FeatureFlag public final boolean useEddReflection;
@@ -623,6 +624,7 @@ public class EDConfig {
     subscriptionSystemActive = getSetupEVBoolean(setup, ev, "subscriptionSystemActive", true);
     convertersActive = getSetupEVBoolean(setup, ev, "convertersActive", true);
     useSaxParser = getSetupEVBoolean(setup, ev, "useSaxParser", false);
+    publishMqttNotif = getSetupEVBoolean(setup, ev, "publishMqttNotif", false);
     enableEnvParsing = getSetupEVBoolean(setup, ev, "enableEnvParsing", true);
     updateSubsRssOnFileChanges = getSetupEVBoolean(setup, ev, "updateSubsRssOnFileChanges", true);
     useEddReflection = getSetupEVBoolean(setup, ev, "useEddReflection", false);


WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
```diff
@@ -193,9 +194,9 @@ public EDDTableFromMqtt(
             sessionExpiryInterval,
             connectionTimeout,
             automaticReconnect);
-    Mqtt5AsyncClient asyncClient = response.join();
+    asyncClient = response.join();
```
ChrisJohnNOAA: This should not be made static. It needs to allow for each dataset to use a different broker.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -24460,6 +24460,14 @@ public static void tryToDoActions(
       try {
         StringArray actions = null;
 
+        // publish change to local broker, if enabled
+        if (EDStatic.config.publishMqttNotif) {
+          // check if client is configured and broker is localhost
+          if (EDDTableFromMqtt.asyncClient != null && EDStatic.config.enableMqttBroker) {
```
ChrisJohnNOAA: Erddap should use it's own client for this broadcast. One that is explicitly pointed at the locally running broker.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
ayushsingh01042003: Not sure I get it, If we dont want static. We will have to call the constructor in tryToDoActions, how will we set the constructor parameters in that case?

ChrisJohnNOAA: We can have a static one stored just for tryToDoActions, but each dataset should have it's own client. The client for tryToDoActions should be in Erddap.java.

The datasets might be connected to different brokers, so they should not use a static client.

ayushsingh01042003: But if thats the case we will have create a client by calling the constructor everytime in somewhere around this code
```
// publish change to local broker, if enabled
        if (EDStatic.config.publishMqttNotif) {
          // check if client is configured and broker is localhost
          if (EDDTableFromMqtt.asyncClient != null && EDStatic.config.enableMqttBroker) {
            EDDTableFromMqtt.publishData(change, "change/" + tDatasetID);
          }
        }
```
That would mean setting the constructor parameters here. Are we going to modify the parser to allow each dataset type config to store it? I still dont get it.

I dont understand what do you mean when saying "static one stored", whats being referred to here?

ChrisJohnNOAA: EDDTableFromMqtt should have a client per instance. Each definition of an EDDTableFromMqtt in datasets.xml needs to be able to specify what broker to connect to. EDDTableFromMqtt should not be involved in the broadcast of dataset changes.

There should be a client (likely saved statically on Erddap.java) that is used for the purpose of broadcasting dataset changes. This would be using the locally running broker. All datasets on the server can share this dataset change broadcast client.

ayushsingh01042003:
Just found that we get the `java.lang.ClassNotFoundException: gov.noaa.pfel.erddap.MqttBrokerContextListener` error on this branch(witout the PR change). Cant seem to figure out the reason.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -159,6 +160,22 @@ public class Erddap extends HttpServlet {
   public static final int dpfTFWidth = 56; // data provider form TextField width
   public static final int dpfTAWidth = 58; // data provider form TextArea width
 
+  // MqttClient to connect to the local ERDDAP broker
+  public static Mqtt5AsyncClient mqttClient =
+      EDDTableFromMqtt.initialiseMqttAsyncClient(
```
ChrisJohnNOAA, line 165: I think the majority of these parameters should be based on the MQTT configuration, not hardcoded.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -159,6 +160,22 @@ public class Erddap extends HttpServlet {
...
+              "localhost",
+              1883,
+              null,
+              "dataset-change-client",
+              "password",
```
ChrisJohnNOAA: We definitely don't want a password hardcoded into code. Can we load this from the MQTT configuration file?

ChrisJohnNOAA (changes requested):

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
ayushsingh01042003: By MQTT configuration file, do you mean setting it in the datasets.xml file and calling the constructor in Erddap.java?
Because the other config.xml is only to set data for the broker

ChrisJohnNOAA: This should be connecting to the locally running broker. The port, username, and password  (and maybe other parameters) should be based on how the MQTT broker is configured. I'm also unsure if "localhost" will work in all circumstances (particularly with some Docker configurations).

ayushsingh01042003: But ther broker is configured by default or by pointing to a config.xml. I dont think the library provides functions to access the attributes of the config files. Will have to write a parser if want to do that

ChrisJohnNOAA: We may need to parse the config file to get that information then.

ChrisJohnNOAA: We could also have config information in the setup.xml for how to connect to the broker. It could default to the values that would most likely work for servers running the embedded broker.

ayushsingh01042003: That does seem like the better approach here. The config file can have bunch of variations so writing a parser might be somewhat of a longer task.

ayushsingh01042003:
Get the following errors on running the jetty server:
```
Caused by: java.lang.NullPointerException: Cannot read field "mqttServerHost" because "gov.noaa.pfel.erddap.util.EDStatic.config" is null
 at gov.noaa.pfel.erddap.Erddap.<clinit>(Erddap.java:164)
```

Followed by exception:
```
Caused by: java.lang.NullPointerException: Cannot read field "mqttServerHost" because "gov.noaa.pfel.erddap.util.EDStatic.config" is null
 at gov.noaa.pfel.erddap.Erddap.<clinit>(Erddap.java:164)
```

Got this problem wen using the hardcoded configurations as well.

Is this something with EDStatic initialization?

ChrisJohnNOAA:
> Get the following errors on running the jetty server:
> 
> ```
> Caused by: java.lang.NullPointerException: Cannot read field "mqttServerHost" because "gov.noaa.pfel.erddap.util.EDStatic.config" is null
>  at gov.noaa.pfel.erddap.Erddap.<clinit>(Erddap.java:164)
> ```
> 
> Followed by exception:
> 
> ```
> jakarta.servlet.ServletException: java.lang.reflect.InvocationTargetException
> ```
> 
> Got this problem wen using the hardcoded configurations as well.
> 
> Is this something with EDStatic initialization?
> 
> Its reproduceable immediately after starting the jetty server

My guess is that there was an exception during initialization of EDStatic which prevented EDstatic.config from being set. Were there any exceptions shortly before this?

WEB-INF/classes/gov/noaa/pfel/erddap/util/EDConfig.java
```diff
@@ -239,6 +253,18 @@ public class EDConfig {
       // used
       useLuceneSearchEngine;
 
+  @FeatureFlag public final String mqttServerHost;
```
ChrisJohnNOAA: FeatureFlag is not appropriate for these configuration values. It is intended to be used on boolean options that enable/disable features in the code.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromMqtt.java
```diff
@@ -302,6 +302,10 @@ public void subscribeToDatasetTopics(Mqtt5AsyncClient client, String[] topics, M
     }
   }
 
+  public void publishData(Mqtt5AsyncClient client, String data, String topic) {
+    client.publishWith().topic(topic).payload(data.getBytes()).send();
```
ChrisJohnNOAA: I don't see this function used now. Remove it?

WEB-INF/classes/gov/noaa/pfel/erddap/util/EDConfig.java
```diff
@@ -239,8 +253,21 @@ public class EDConfig {
...
+  @FeatureFlag public final boolean mqttSsl;
+  @FeatureFlag public final int mqttKeepAlive;
+  @FeatureFlag public final boolean mqttCleanStart;
+  @FeatureFlag public final int mqttSessionExpiry;
+  @FeatureFlag public final int mqttConnectionTimeout;
+  @FeatureFlag public final boolean mqttAutomaticReconnect;
+
   @FeatureFlag public final boolean variablesMustHaveIoosCategory;
   @FeatureFlag public boolean useSaxParser;
+  @FeatureFlag public boolean publishMqttNotif;
```
ChrisJohnNOAA, line 270: Just to make sure its clear. This should use @FeatureFlag.

WEB-INF/classes/gov/noaa/pfel/erddap/Erddap.java
```diff
@@ -24460,6 +24477,11 @@ public static void tryToDoActions(
       try {
         StringArray actions = null;
 
+        // publish change to local broker, if enabled
+        if (EDStatic.config.publishMqttNotif && EDStatic.config.enableMqttBroker) {
```
ChrisJohnNOAA: Since we allow users to configure the notification broker now, maybe we get rid of the enableMqttBroker check here?

ChrisJohnNOAA (approved):

PR #352, opened 2025-08-29
Jte gsoc3 shane updates

srstsavage:
* Restore selective supported html layout page logic from harish-2003/erddap#1


srstsavage:
Haha oops.

PR #369, opened 2025-09-26
Handle some EDDTableFromMqtt edge cases

ChrisJohnNOAA:
# Description

If there's no data file at startup, create an empty file.
Properly save file table changes  with MQTT updates (in both in memory and on disk modes).
Add a helper script that can send an mqtt message (useful for adding data to a locally running server).

Fixes # (issue)

## Type of change

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromFiles.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDTableFromFiles.java
@@ -1387,35 +1387,6 @@ public abstract class EDDTableFromFiles extends EDDTable implements WatchUpdateH
     cachePartialPathRegex =
         String2.isSomething(tCachePartialPathRegex) ? tCachePartialPathRegex : null;
 
-    // class-specific things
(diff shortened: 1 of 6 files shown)


ayushsingh01042003:
I checked out this branch and noticed that I was not setting the data folder in the setup.xml once I did. The dataset started to load. This was the reason for it to fail at the presentation too..

So there was no problem with the application. Unless if we want to handle this case of what to do if the user does not setup their data directory?

ChrisJohnNOAA:
> I checked out this branch and noticed that I was not setting the data folder in the setup.xml once I did. The dataset started to load. This was the reason for it to fail at the presentation too..
> 
> So there was no problem with the application. Unless if we want to handle this case of what to do if the user does not setup their data directory?

This does handle an empty data directory, but it also improves the saving of the filetable and dirtable. I think both are worthwhile improvements.

Also, there are likely to be many users that want to use this to collect MQTT messages and don't already have a datafile. I expect by handling the case where there is no data file we will prevent the most common question users have trying to set up an EDDTableFromMqtt dataset.

PR #373, opened 2025-10-08
Add Pressure support to EDVAlt

SethChampagneNRL:
# Description

Adds a special case for elevations defined by pressure. This is primarily used in Meteorology datasets defining vertical elevations in isobaric levels. NOTE: Smaller pressure values mean higher elevations, so the axis runs opposite the normal elevations defined in meters or feet. 

## Type of change

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGrid.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGrid.java
@@ -202,8 +202,9 @@ public abstract class EDDGrid extends EDD {
     } else if (EDV.ALT_NAME.equals(tDestName)) {
       if (av >= 0) altIndex = av;
       return new EDVAltGridAxis(
-          tParentDatasetID, tSourceName, tSourceAtt, tAddAtt, sourceAxisValues);
-    } else if (EDV.PRESSURE_LONGNAME.equals(ioosCategory)) {
+          tParentDatasetID, tSourceName, tSourceAtt, tAddAtt, sourceAxisValues, false);
+    } else if (EDV.PRESSURE_NAME.equals(tDestName) || EDV.PRESSURE_LONGNAME.equals(ioosCategory)) {
+      // destinationName = "pressure" or ioosCategory = "Pressure" are supported for isobaric levels
       if (av >= 0) altIndex = av;
       return new EDVAltGridAxis(
           tParentDatasetID, tSourceName, tSourceAtt, tAddAtt, sourceAxisValues, true);
--- a/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGridFromErddap.java
+++ b/WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGridFromErddap.java
@@ -445,7 +445,7 @@ public class EDDGridFromErddap extends EDDGrid implements FromErddap {
       switch (tName) {
         case EDV.LON_NAME -> lonIndex = av;
         case EDV.LAT_NAME -> latIndex = av;
-        case EDV.ALT_NAME -> altIndex = av;
+        case EDV.ALT_NAME, EDV.PRESSURE_NAME -> altIndex = av;
         case EDV.DEPTH_NAME -> depthIndex = av;
         case EDV.TIME_NAME -> timeIndex = av;
       }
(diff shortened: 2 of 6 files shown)


WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGrid.java
```diff
@@ -196,6 +203,10 @@ public EDVGridAxis makeAxisVariable(
       if (av >= 0) altIndex = av;
       return new EDVAltGridAxis(
           tParentDatasetID, tSourceName, tSourceAtt, tAddAtt, sourceAxisValues);
+    } else if (EDV.PRESSURE_LONGNAME.equals(ioosCategory)) {
```
ChrisJohnNOAA: Can we add a comment for why this checks ioosCategory when the other checks here use destName.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGridFromErddap.java
```diff
@@ -449,6 +449,8 @@ else if (EDVTime.hasTimeUnits(language, tSourceAttributes, tAddAttributes))
         case EDV.DEPTH_NAME -> depthIndex = av;
         case EDV.TIME_NAME -> timeIndex = av;
       }
+      String ioosCategory = axisVariables[av].sourceAttributes().getString("ioos_category");
```
ChrisJohnNOAA, line 452: Can we add a comment for why this checks ioos_category when the other checks here use destinationName.

WEB-INF/classes/gov/noaa/pfel/erddap/variable/EDVAltGridAxis.java
```diff
@@ -73,6 +73,67 @@ public EDVAltGridAxis(
...
+   * @param isPressure boolean flag used to run this constructor to build pressure axis variable
+   * @throws Throwable if trouble
+   */
+  public EDVAltGridAxis(
+      String tParentDatasetID,
+      String tSourceName,
+      Attributes tSourceAttributes,
+      LocalizedAttributes tAddAttributes,
+      PrimitiveArray tSourceValues,
+      boolean isPressure)
```
ChrisJohnNOAA, line 42: I'm not a fan of adding an unused variable to differentiate constructors. It looks like the existing constructor is only called in one spot, so it should be easy to update the existing call site. Could you do one of the following:

1) Replace the public constructors with static create methods. This would allow naming them differently even if they took the same parameters.
2) Having a single constructor with the isPressure boolean used for the logic differences between the two (they do look quite similar).

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGrid.java
SethChampagneNRL: I had copied this change from another developer. It was part of NAVO's fork. I think the reasoning is most of the Meteorology data is configured with the axis named "isobaric" or "isbr_lvl", or some other variation. 

We could either leave this as is using the IOOS Category, or we could enforce these axes set a standard destination name like "pressure". If we did that, then it would behave similar to altitude and depth.

WEB-INF/classes/gov/noaa/pfel/erddap/variable/EDVAltGridAxis.java
SethChampagneNRL: I will switch it to option 2. I think this was a copy/paste mistake on my part.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGridFromErddap.java
SethChampagneNRL: Same as above.

WEB-INF/classes/gov/noaa/pfel/erddap/dataset/EDDGrid.java
ChrisJohnNOAA: I'd prefer to be as consistent as possible with the [special variables](https://erddap.github.io/docs/server-admin/datasets?_highlight=altitude#special-variables). That said I could see an argument that it would be good to support ioos category OR destination name at least for pressure (and maybe time).

SethChampagneNRL: Well actually, we have another use case where we have two Time dimensions, and we set both axes to IOOS Category "Time" (analysis time and forecast time, basically having overlapping forecasts loaded into a single dataset). So personally I prefer the special case destination name.

ChrisJohnNOAA: How about we support ioos category or destination name for pressure and make an agenda item for the next technical board to get input on other changes?

SethChampagneNRL: Good plan.

PR #374, opened 2025-10-09
Bugfix/allow delete char

SethChampagneNRL:
# Description

We encountered an issue with some NetCDF files that include the DEL char (code 127) in one of their source attributes. The test for printable characters would reject these NetCDF files and cause the dataset to not load with a confusing error log. Allowing the DEL char does not break any of the ERDDAP functionality for the griddap dataset. 

In my opinion, this was an overly restrictive test. This is the simplest code fix I could find without requiring editing the NetCDF data. 

## Type of change

### DIFF
--- a/WEB-INF/classes/com/cohort/util/String2.java
+++ b/WEB-INF/classes/com/cohort/util/String2.java
@@ -4974,7 +4974,7 @@ public class String2 {
 
   /**
    * This returns the index of the first non-Unicode character. Currently, valid characters are #32
-   * - #126, #160+.
+   * - #127, #160+.
    *
    * @param s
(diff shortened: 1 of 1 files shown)


WEB-INF/classes/com/cohort/util/String2.java
```diff
@@ -4986,7 +4986,7 @@ public static int findInvalidUnicode(String s, String alsoOK) {
       char ch = s.charAt(i);
       if (alsoOK.indexOf(ch) >= 0) continue;
       if (ch < 32) return i;
-      if (ch <= 126) continue;
+      if (ch <= 127) continue; // allow up to DELETE char
```
ChrisJohnNOAA, line 4989: Please also update the comment for this function.

Issue #311, opened 2025-06-24
S3 aggregations do not properly descend prefixes


benjwadams:
**Describe the bug**
S3 EDDTableFromDatasets
**To Reproduce**
Create subdirectories with the following scheme:
DEPLOYMENT_NAME
   - 2025-05
   - 2025-06

The month subfolders should both contain files.

Recreate the hierarchy in AWS S3 using `aws cp` or a similar tool.
**Expected behavior**
Aggregations have the same data in both the local filesystem versions and S3 version.

Actual behavior shows only one month of data in the S3 version.
**Screenshots**

**Server**
- What ERDDAP™ server? If you're the admin, please provide what version of ERDDAP™, Tomcat, and Java you are using.

2.25.1, Axiom Docker image

**Desktop (please complete the following information):**
 - OS: Linux
**Additional context**

Abridged original email sent to @ChrisJohnNOAA:

```Hi,

   Following up on some ERDDAP S3 issues, I noticed that prefixes are not picked up on S3 datasets in the same way that subdirectories are picked up

Consider the following directory structure:

WOOD/
     2025-05/
     2025-06/

EDDFromAsciiFiles picks these up from both subdirectories when from the local system, but not when using files from S3.  Compare:

(local — disregard name, 2025-05 and 2025-06)
[ERDDAP - radials/WOOD data from AWS S3 bucket rps-nccf-hfrnet-processed-uat - Data Access Form](https://hfradar.ioos.us/dev/erddap/tabledap/WOOD_local_files.html)
(S3 based, 2025-05 only)
[ERDDAP - radials/WOOD data from AWS S3 bucket rps-nccf-hfrnet-processed-uat - Data Access Form](https://hfradar.ioos.us/dev/erddap/tabledap/WOOD.html)
```

Attached is also a minimal datasets.xml file with a changed extension to pass through GitHub.  Currently the bucket is public read so this should be reproducible on the existing setup once the files are downloaded.

[datasets_abridged.xml.txt](https://github.com/user-attachments/files/20889651/datasets_abridged.xml.txt)

ChrisJohnNOAA:
I believe this is resolved with @SethChampagneNRL 's changes. Marking as closed. If anybody still has the issue on the latest nightly build or in 2.29.0 (or later) once it's released, please reopen.

Issue #302, opened 2025-06-09
AWS S3 datasets leak threads, also when under load, leaks memory


ChrisJohnNOAA:
**Describe the bug**

There is evidence that specifically when pulling datasets from S3 there are threads that don't get cleaned up and memory when the server is trying to load too many datasets. It was discussed and described in this thread on the google group.

https://groups.google.com/g/erddap/c/BHdELzgWf7o

Based on that conversation, I suspect the AWS integration is leaving threads around indefinitely. Also, if the datasets aren't completed loading when the next load datasets starts, it might be leaking threads with significant memory usage.

**Expected behavior**

Threads get cleaned up after use and no memory growth.

**Server**
- What ERDDAP™ server? If you're the admin, please provide what version of ERDDAP™, Tomcat, and Java you are using.

I believe the server is this one: https://seagull-erddap.glos.org/erddap/
ERDDAP version 2.26, using Docker
Java: 21.0.6+7-LTS

ChrisJohnNOAA:
I believe this is resolved with @SethChampagneNRL 's changes. Marking as closed. If anybody still has the issue on the latest nightly build or in 2.29.0 (or later) once it's released, please reopen.

joe-smithe-glos:
Thanks so much. Will keep you posted!

-Joe

Joeseph Smith

Cyberinfrastructure Engineer

734-972-8096

*www.glos.org <http://www.glos.us/>*

LinkedIn <https://www.linkedin.com/company/great-lakes-observing-system/> |
Facebook <https://www.facebook.com/GreatLakesObservingSystem/> | Instagram
<https://www.instagram.com/greatlakesobservingsystem/> | YouTube
***@***.***>

Issue #107, opened 2023-08-23
Overly restrictive S3 limitations


benjwadams:
I've recently been using ERDDAP, version 2.23 using Axiom's Docker images in an attempt to load in HF Radar files with the data sections stored as CSV which are stored in S3.
Bucket is here and is currently publicly readable:

https://hfradar.s3.amazonaws.com/

I've run into several surprising behaviors.

1)  ERDDAP docs do mention to use a particular availablility zone, e.g. instead of using https://hfradar.s3.amazonaws.com/SUBPATH/, use https://hfradar.s3.us-east-1.amazonaws.com/SUBPATH/ .   However, the first form is usually what you'd get from copying from the AWS console, and also would support multiple availability zones.   From all I can determine, the content served by both is the same barring downtime error conditions, but the former has the advantage of multi-AZ support.  ERDDAP wants the other form and if you do not provide it, you will get an error message saying that no files were found.  Docs do say to use a form with AZ specified, but this is confusing behavior given that the content is going to be the same in 99.9%+ of cases.  I don't understand why the former form can't be supported, especially given that it's likely to be more robust in a production environment given the fact that multiple availability zones are supported.

2) Even though the bucket and its contents are publicly readable, ERDDAP will refuse to read the contents of the S3 bucket if you don't pass along credentials.  Why is this step necessary for something that should already be readable without supplying AWS credentials?

benjwadams:
Confirmation that both HTTPS URL forms yield the same result:

```
[badams@localhost]~% curl -s https://hfradar.s3.us-east-1.amazonaws.com/ | md5sum
e3d288cb10af4d93a3509f3776a45dec  -
[badams@localhost]~% curl -s https://hfradar.s3.amazonaws.com/ | md5sum       
e3d288cb10af4d93a3509f3776a45dec 
```

BobSimons:
1) When I updated to v2 of the AWS Java SDK, the only way I could get it to work was to specify the zone, so I chose the form of the URL which includes the zone. (Note that when just using the URL, e.g., in a browser, I didn't need to specify the zone. The SDK works differently.) Note that the ERDDAP code doesn't simply use this URL, it parses it and uses the pieces, including the zone.   I didn't know about the multiAZ option. Perhaps there is a way to support this with the SDK. That would be up to Chris to pursue if he chooses to. 

[...]

PR #375, opened 2025-10-09
Support for S3 URI

### LINKED ISSUE #95: S3 Support
Hello,

We've recently looked into using our ERRDAP instance with a Minio S3 object store. Unfortunately it looks like ERDDAP currently only works with AWS S3 locations. Is this something that could be added in? It certainly would be a big help for us and hopefully could be useful for others

SethChampagneNRL:
# Description

Adding support for S3 URIs as the `cacheFromUrl` value. This will allow ERDDAP to support private buckets hosted off amazonaws.com

Also addressing the memory leak issue. I believe it is caused by the S3Clients not getting closed.

Fixes #95 #107 #302 #311 

## Type of change

### DIFF
--- a/WEB-INF/classes/gov/noaa/pfel/coastwatch/util/SSR.java
+++ b/WEB-INF/classes/gov/noaa/pfel/coastwatch/util/SSR.java
@@ -910,7 +910,6 @@ public class SSR {
               .credentialsProvider(credentialsProvider)
               .region(Region.of(region))
               .targetThroughputInGbps(20.0) // ??? make a separate setting?
-              .targetThroughputInGbps(20.0)
               .minimumPartSizeInBytes((long) (8 * Math2.BytesPerMB));
     } else {
       builder =
--- a/src/test/java/gov/noaa/pfel/coastwatch/util/FileVisitorDNLSTests.java
+++ b/src/test/java/gov/noaa/pfel/coastwatch/util/FileVisitorDNLSTests.java
@@ -518,15 +518,7 @@ class FileVisitorDNLSTests {
     String2.log("\n*** FileVisitorDNLS.testAWSS3 finished.");
   }
 
-  /**
-   * This tests this class with Amazon AWS S3 file system. Your S3 credentials must be in <br>
-   * ~/.aws/credentials on Linux, OS X, or Unix <br>
-   * C:\Users\USERNAME\.aws\credentials on Windows See
-   * https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html See
-   * https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html See
-   * https://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html See
-   * https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/setup.html#setup-credentials
-   */
+  /** This tests this class with Amazon AWS S3 file system using the */
   @org.junit.jupiter.api.Test
   @TagAWS
   void testAWSS3WithS3URI() throws Throwable {


WEB-INF/classes/com/cohort/util/File2.java
```diff
@@ -1138,31 +1143,14 @@ public static boolean isDecompressible(String ext) {
...
-      s3Client =
-          S3Client.builder()
-              // was .credentials(ProfileCredentialsProvider.create()) //now it uses default
-              // credentials
-              .region(Region.of(region))
-              .build();
-      s3ClientMap.put(region, s3Client);
-    }
-    return s3Client;
+    // FIXME close all instances
```
ChrisJohnNOAA: I think they're all closed explicitly now except for the code path that wraps the S3Client in a BufferedInputStream. That should close the client when the input stream is closed.

ChrisJohnNOAA: ErrorProne does throw an error because of the s3Client in the getBufferedInputStream method at line 1197 in here. Since the s3Client is getting wrapped in there, it might need to be annotated with @SuppressWarnings("MustBeClosedChecker") (it's possible adding @MustBeClosed would work and push this to the next layer, but I don't know).

SethChampagneNRL: Yesterday I got this branch working with our self-hosted S3. There's still a variety of manual testing I want to try today, and I will try to put in some unit test cases based on the existing tests but using the s3:// URI schema. 

I will also update the docs with how to set up an S3 custom endpoint using the system properties provided by the AWS SDK. 

The `BufferedInputStream` case for the `S3Client` was the only case that I knew was going to be tricky. I need to figure out how to trigger that code in my debugger and see if it works as expected. Otherwise I will have to jump through some hoops to make sure it is closed. If what you said above turns out to be true, I will suppress that warning.

SethChampagneNRL: I was able to hit this chunk of code through the "files" download links. The `S3Client` and `InputStream` were not closed together, and closing the client prematurely leaves the stream unable to read the data. 

My solution was to wrap it in the `ProxyInputStream` from commons-io, and override its `close()` method to close the S3Client.

WEB-INF/classes/gov/noaa/pfel/coastwatch/util/SSR.java
```diff
@@ -894,16 +899,26 @@ public static BufferedInputStream getUncompressedUrlBufferedInputStream(String u
...
+    if (EDStatic.config.useAwsAnonymous) {
+      credentialsProvider = AnonymousCredentialsProvider.create();
+    }
+    if (EDStatic.config.useAwsCrt) {
+      builder =
+          S3AsyncClient.crtBuilder()
+              .credentialsProvider(credentialsProvider)
+              .region(Region.of(region))
+              .targetThroughputInGbps(20.0) // ??? make a separate setting?
+              .targetThroughputInGbps(20.0)
```
ChrisJohnNOAA: targetThroughputInGbps is set twice here, and not set in the else branch. Is that intended?

Also should minimumPartSizeInBytes be in the else?

src/test/java/gov/noaa/pfel/coastwatch/util/FileVisitorDNLSTests.java
```diff
@@ -518,6 +518,171 @@ void testAWSS3() throws Throwable {
     String2.log("\n*** FileVisitorDNLS.testAWSS3 finished.");
   }
 
+  /**
+   * This tests this class with Amazon AWS S3 file system. Your S3 credentials must be in <br>
```
ChrisJohnNOAA: If we have useAwsAnonymous = true do we need credentials for these tests?

ChrisJohnNOAA (changes requested):
I'm excited for this change. Thanks for working on it,

One small change and one clarifying question.

WEB-INF/classes/gov/noaa/pfel/coastwatch/util/SSR.java
SethChampagneNRL: The `targetThroughputInGbps` looks unintentional. I will remove the extra call. 

I will take a look at `minimumPartSizeInBytes`.

src/test/java/gov/noaa/pfel/coastwatch/util/FileVisitorDNLSTests.java
SethChampagneNRL: These tests will work with the anonymous credentials. I forgot to update the javadoc.

WEB-INF/classes/gov/noaa/pfel/coastwatch/util/SSR.java
SethChampagneNRL: Turns out `minimumPartSizeInBytes` is an AWS CRT only method.

ChrisJohnNOAA (approved):

PR #379, opened 2025-10-20
Add support for EDDGridLonPM180 to fix longitudes greater than 360

SethChampagneNRL:
# Description

I am adding support for ECMWF Grib files pulled form their Python API https://pypi.org/project/ecmwf-opendata/

These Grib files have longitude in the range 180 to 540. I would like to correct these into -180 to 180. The basic logic for this was already supported by `EDDGridLonPM180`, but I needed to increase the maximum value from 360 to 540. Adding an optional dataset parameter did the trick. 

```
 <dataset type="EDDGridLonPM180" datasetID="ECMWF-FIXED" active="true">
    <maxSourceLon>540</maxSourceLon>
    <dataset type="EDDGridFromNcFiles" datasetID="ECMWF" active="true">
        <!-- redacted internals -->
    </dataset>
</dataset>
```

Another issue in the ECMWF dataset is the latitudes running from 90 to -90 instead of -90 to 90. I would like to also correct these, but I have not figured out a solution yet. 

## Type of change


rmendels:
@SethChampagneNRL   @ChrisJohnNOAA   you said "Another issue in the ECMWF dataset is the latitudes running from 90 to -90 instead of -90 to 90. I would like to also correct these, but I have not figured out a solution yet."  There are existing netcdf datasets that have latitudes running from 90 to -90  so there must be code already there to do what you want.

ChrisJohnNOAA:
> latitudes running from 90 to -90 instead of -90 to 90

From the documentation: https://erddap.github.io/docs/server-admin/datasets?_highlight=latitude#special-variables

> For the "longitude" variable and the "latitude" variable:
> Use the [destinationName](https://erddap.github.io/docs/server-admin/datasets?_highlight=latitude#destinationname)s "longitude" and "latitude" only if the [units](https://erddap.github.io/docs/server-admin/datasets?_highlight=latitude#units) are degrees_east and degrees_north, respectively. If your data doesn't fit these requirements, use different variable names (for example, x, y, lonRadians, latRadians).
> If you have longitude and latitude data expressed in different units and thus with different destinationNames, for example, lonRadians and latRadians, Make A Graph (datasetID.graph) will make graphs (for example, time series) instead of maps.

Are the datasets that currently do it using a destination variable named "latitude" or do they call it something else?

You might be able to use [scale_factor](https://erddap.github.io/docs/server-admin/datasets?_highlight=scale_factor#scale_factor) to adjust the latitude.

`<att name="scale_factor" type="float">-1</att>`

rmendels:
@SethChampagneNRL  @ChrisJohnNOAA You can just use latitude and degrees_north.  Here is one example:

https://coastwatch.pfeg.noaa.gov/erddap/griddap/erdVHsstaWS3day.html

In fact I think most if not all the VIIRS datasets go north to south, and there are others but I am too lazy to find them.  I don't know what ERDDAP does internally but it doesn't seem to be bothered by it.

src/test/java/testDataset/EDDTestDataset.java
```diff
@@ -34194,6 +34195,198 @@ private static String xmlFragment_notApplicable_LonPM180() {
...
   }
 
+  private static String xmlFragment_ecmwfTest_LonPM180() {
+    return """
+    <dataset type="EDDGridLonPM180" datasetID="ECMWF-FIXED" active="true">
+        <maxSourceLon>540</maxSourceLon>
+        <dataset type="EDDGridFromNcFiles" datasetID="ECMWF" active="true">
+            <reloadEveryNMinutes>10080</reloadEveryNMinutes>
+            <updateEveryNMillis>10000</updateEveryNMillis>
+            <fileDir>C:\\Users\\schampagne\\Workspace\\nerddap\\erddap\\test-data\\data</fileDir>
```
ChrisJohnNOAA: This should use a resource, for example:
Path.of(EDDTestDataset.class.getResource("/data/").toURI()).toString()

That allows project relative loading of data files.

src/test/java/gov/noaa/pfel/erddap/dataset/EDDGridLonPM180Tests.java
```diff
@@ -637,6 +637,103 @@ void test0to360() throws Throwable {
...
   }
 
+  /**
+   * This tests a dataset that is initially 180to540.
+   *
+   * @throws Throwable if trouble
+   */
+  @org.junit.jupiter.api.Test
+  @TagMissingDataset // Souce 404
+  @TagImageComparison
```
ChrisJohnNOAA: I don't see an image comparison in the test. If there is not an image comparison this tag shouldn't be used.

src/test/java/gov/noaa/pfel/erddap/dataset/EDDGridLonPM180Tests.java
```diff
@@ -637,6 +637,103 @@ void test0to360() throws Throwable {
     // String2.log("\n*** EDDGridLonPM180.test0to360 finished.");
   }
 
+  /**
+   * This tests a dataset that is initially 180to540.
+   *
+   * @throws Throwable if trouble
+   */
+  @org.junit.jupiter.api.Test
+  @TagMissingDataset // Souce 404
```
ChrisJohnNOAA: This tag is used for old tests that have datasets that are not currently accessible. I'd prefer to not add a test using the tag. If the test data is added in /src/test/resources or to a new version of the test data zip files, we should be able to run the test for everyone.

ChrisJohnNOAA (changes requested):

src/test/java/testDataset/EDDTestDataset.java
SethChampagneNRL: Oops, forgot to switch back from my hardcode URL. Will fix.

SethChampagneNRL:
I think I can download a lower resolution grib and make it smaller. But it should still go to the zip file instead of erddap's repo imo.

ChrisJohnNOAA:
> Never mind, ECMWF only offers 0.25 resolution.

I made a new release of the test data. The pom for this pull needs to be updated to use test1.05

ChrisJohnNOAA (approved):

PR #391, opened 2025-11-03
Some optimizations and speeding up tests

ChrisJohnNOAA:
# Description

This includes several commits that are kept separate on purpose to make it easier to review.
- Two passes at reducing/removing explicit sleep calls during tests.
- Refactor to remove the StringHolder class.
- Migration to using java.nio.OutputStream.
- Multiple optimizations to String handling functions in order to reduce allocations during processing. This was largely focused on Table.readAscii performance, but many other parts of the code will get the improvements as well.
- Move to the modern java.nio.File move call (instead of the old complicated file rename approach. 
- Separate out tests that are really testing external servers and not testing the code base. These are very flaky and don't provide useful signal during development.

## Type of change


srstsavage:
Great changes! Unit test time is getting much more manageable.

PR #453, opened 2026-04-08
Update version to 2.31.0-alpha for next dev cycle/dependency updates

srstsavage:
# Description

Update version in pom.xml and EDStatic.java to 2.31.0-alpha for next development cycle.

Also perform minor upgrades on dependencies.


rmendels (approved):
I have reviewed these changes, and approve them.

PR #510, opened 2026-07-08
Create a JSON webservice for ERDDAP's variable attribute suggestions

srstsavage:
# Description

Provide ERDDAP's variable attribute suggestions in EDD.makeReadyToUseAddVariableAttributesForDatasetsXml as a JSON webservice, accepting an ncoJson style variable attribute payload and returning suggested attributes (standard name, long name, etc).

This is useful for external tools building datasets.xml config or otherwise curating variable attributes.

## Type of change

### DIFF
--- a/WEB-INF/classes/com/cohort/array/Attributes.java
+++ b/WEB-INF/classes/com/cohort/array/Attributes.java
@@ -920,10 +920,12 @@ public class Attributes {
    *
    * @param indent a String a spaces for the start of each line
    * @param includeTrailingComma if true, a comma is included after the closing } of the attributes
+   * @param includeWrapper if true, the attributes are wrapped in "attributes": { ... }
    * @return a string with all of the attributes for a variable (or *GLOBAL*) formatted for NCO
    *     JSON, with a comma after the closing }.
    */
-  public String toNcoJsonString(String indent, boolean includeTrailingComma) {
+  public String toNcoJsonString(
+      String indent, boolean includeTrailingComma, boolean includeWrapper) {
     StringBuilder sb = new StringBuilder();
 
     // "attributes": {
@@ -938,7 +940,10 @@ public class Attributes {
     //  "double_att": { "type": "double", "data": [73.0, 72.0, 71.0, 70.010, 69.0010, 68.010,
     // 67.0100010]}
     // },
-    sb.append(indent + "\"attributes\": {\n");
+    if (includeWrapper) {
+      sb.append(indent + "\"attributes\": ");
+    }
+    sb.append("{\n");
 
     // each of the attributes
     String names[] = getNames();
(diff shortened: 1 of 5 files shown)


srstsavage:
Some review considerations here:

* This new service currently lives at `/erddap/suggestVariableAttributes`. We could rehome it under a new ERDDAP protocol like `/erddap/services/suggestVariableAttributes` or something similar if desired. `/erddap/rest` is unfortunately claimed by ESRI REST services.
* Speaking of ESRI REST, `doSuggestVariableAttributes` is currently co-opting `sendGeoServicesRestError` as a simple way to return JSON error messages. I also updated `sendGeoServicesRestError` to set the "application/json" content header (`.json` to `getJsonWriter` instead of `.jsonFile`) and also set the HTTP status code correctly. I believe both of those things weren't previous done due to some [incorrect behavior by ESRI](https://community.esri.com/t5/arcgis-rest-apis-and-services-questions/why-aren-t-rest-json-responses-using-content-type/td-p/159566) which has since been fixed). If this isn't desired it would be easy to create a seperate JSON error handler and leave `sendGeoServicesRestError` alone.
* For simplicity, `doSuggestVariableAttributes` is using `response.getWriter()` and setting Content-Type as needed. Could move to another writer if needed.
* Error message are not currently translated, not sure if that's a requirement for a simple JSON webservice.

ChrisJohnNOAA (approved):

ChrisJohnNOAA:
> Some review considerations here:
> 
> * This new service currently lives at `/erddap/suggestVariableAttributes`. We could rehome it under a new ERDDAP protocol like `/erddap/services/suggestVariableAttributes` or something similar if desired. `/erddap/rest` is unfortunately claimed by ESRI REST services.
> * Speaking of ESRI REST, `doSuggestVariableAttributes` is currently co-opting `sendGeoServicesRestError` as a simple way to return JSON error messages. I also updated `sendGeoServicesRestError` to set the "application/json" content header (`.json` to `getJsonWriter` instead of `.jsonFile`) and also set the HTTP status code correctly. I believe both of those things weren't previous done due to some [incorrect behavior by ESRI](https://community.esri.com/t5/arcgis-rest-apis-and-services-questions/why-aren-t-rest-json-responses-using-content-type/td-p/159566) which has since been fixed). If this isn't desired it would be easy to create a seperate JSON error handler and leave `sendGeoServicesRestError` alone.
> * For simplicity, `doSuggestVariableAttributes` is using `response.getWriter()` and setting Content-Type as needed. Could move to another writer if needed.
> * Error message are not currently translated, not sure if that's a requirement for a simple JSON webservice.

I don't believe we're consistent on translating error messages. It's good to, but not required.

srstsavage:
Sounds good, I'll leave as-is for now then.

You're fine with the URL of `/erddap/suggestVariableAttributes`? No need to create a new `/erddap/services/` route?

ChrisJohnNOAA:
> Sounds good, I'll leave as-is for now then.
> 
> You're fine with the URL of `/erddap/suggestVariableAttributes`? No need to create a new `/erddap/services/` route?

I think its good for now. We might change later when we have a better idea on offering a set of machine friendly services.

PR #513, opened 2026-07-14
Replace URL detection with autolink-java

ChrisJohnNOAA:
# Description

I've started using CodeQL to highlight code concerns (it has a focus on security and performance issues). CodeQL highlighted the url extraction regexes as potentially extremely slow due to backtracking. While looking into fixes, I came across a library that does a better job than the regexes (there's some edge cases it handles correctly) and doesn't trigger CodeQL warnings.

Since I don't believe we have legitimate performance concerns with either approach I did not run a performance test.

Fixes CodeQL warnings. These aren't enabled on the main ERDDAP repo yet, and likely show up as security issues so I don't think I could link to them.

## Type of change


WEB-INF/classes/com/cohort/util/LinkHelper.java
```diff
@@ -0,0 +1,182 @@
+package com.cohort.util;
+
+import java.util.EnumSet;
+import org.nibor.autolink.LinkExtractor;
```
srstsavage, line 4: Love the isolation of the autolink dependency to this `LinkHelper` class in case it ever needs to be replaced.

srstsavage (approved):

PR #593, opened 2026-09-18
cache the S3TransferManager and cleanup on EDStatic.destroy()

### LINKED ISSUE #592: S3TransferManager causing issues?
Working with ERDDAP to provide file-based access to our data between an s3 bucket and ERDDAP's s3Cache. Additionally working with Claude to ensure we have everything setup correctly and leverage what's in the ERDDAP code base. Ran into an interesting issue which may be a follow up to #302 . Claudian output follows. Feel free to ignore or delete if need be.

## Summary

`SSR.downloadFile()` opens a brand-new `S3TransferManager` — and therefore a brand-new native AWS CRT S3 client with its own OS thread pool — for **every individual file** it downloads from S3, via `ensureInCache()`'s download path (used by `cacheFromUrl`/bounded-cache datasets, e.g. `EDDTableFromFiles` subclasses with `cacheFromUrl` + `cacheSizeGB` set).

This is a follow-up to #302, not a fresh, unrelated bug — see "Relationship to #302" below.

## Impact

Any dataset that needs to harvest metadata from many remote files in one construction/reload pass — for example, a multi-year archive stored as one file per day, hundreds of files total — creates and tears down hundreds of `S3TransferManager`/CRT client instances in quick succession during a single dataset load.

We instrumented the actual JVM process's thread count (`/proc/<pid>/status`, `Threads:`) while a single dataset with several hundred remote files was being (re)constructed, and observed thread count growing roughly linearly with files processed so far, not stabilizing around a steady-state working set:

```
54 → 1,192 → 3,705 → 6,266 → 8,855 → 11,417 → 14,042 → 16,730 → 19,122 →
21,229 → 22,957 → 24,722 → 26,781 → 29,387 → 31,976 → ...
```

(roughly +2,000–2,600 threads per 30 seconds, still climbing when we stopped observing). Cleanup from a closed `S3TransferManager` is not keeping pace with creation of the next one.

Depending on the container/host's resource limits, this shows up as one of two failure modes:

1. **Thread/process limit exhaustion.** Once the OS-level ceiling is hit, new CRT client creation fails outright:
   ```
   java.io.IOException: ERROR in ensureInCache (AWS Transfer Manager) while downloading from <url>
    at gov.noaa.pfel.coastwatch.util.SSR.downloadFile(SSR.java:1048)
   Caused by: software.amazon.awssdk.crt.CrtRuntimeException: S3Client.aws_s3_client_new:
   creating aws_s3_client failed (aws_last_error: AWS_ERROR_THREAD_INSUFFICIENT_RESOURCE(12),
   Insufficient resources for thread.)
   ```
   Each failure drops that file into the standard "bad file" handling (`badFiles.nc`), and — because immutable archival files never change `lastModified` — it's never retried on a later reload. In one test, this affected roughly a quarter of the files in the dataset.

2. **Out-of-memory.** With a more generous thread/process limit, the same growth appears to drive genuine OOM kills instead — observed even on a container with tens of GB of memory headroom, for a dataset whose actual file payload was a small fraction of that. Points to native/off-heap memory tied to repeated client allocation/teardown, not heap pressure from the downloaded data.

Raising the container's thread/process ulimit only delays the problem — it shifts which failure mode you hit and when, since the growth doesn't appear to plateau within a single construction pass.

## Relationship to #302

This is the unresolved half of #302, not a new bug. The fix that shipped for #302 (`80bd2f11`, "allow S3 URIs as cacheFromUrl, close S3Clients") touches this exact method:

```diff
-      try {
-        S3TransferManager tm = buildS3TransferManager(bro[1]); // !!! ??? reuse these???
+      try (S3TransferManager tm = buildS3TransferManager(bro[1]); ) {
```

Before that fix, the `S3TransferManager` was never closed at all — an unconditional, permanent leak, one client per download, forever, for the life of the JVM. The fix wrapped the call in `try-with-resources`, so the client is now closed after each download. That's a real fix for the *leak* half of the problem.

What it didn't do — flagged in the original developer's own inline comment on the leaking line, `// !!! ??? reuse these???` — is **reuse** a client across calls. The fix that shipped is "create one per file, then close it," not "share one across files." That gap is exactly what this issue is about: closing promptly isn't the same as closing *fast enough*. Under the volume of a real multi-year archive (hundreds of sequential create+close cycles in one construction pass), teardown of each native CRT thread pool doesn't keep pace with creation of the next one, so thread count still climbs — just via "leak faster than cleanup" instead of "never cleaned up at all."

## Suggested fix

Reuse a single `S3TransferManager` (and its underlying CRT client) across all downloads for a given bucket/region within one calling operation (e.g., one per dataset construction/reload, or a small pool keyed by region), rather than constructing and closing one per file — i.e. what the original `// !!! ??? reuse these???` comment already called for.

Pressing Claude as to why using a single `S3TransferManager` wasn't the immediate goto, it claimed "current code has no scaffolding for it":

> 1. buildS3TransferManager(region) is a bare static factory, no cache/registry. It just builds and returns a new S3TransferManager every call. There's no Map<region, S3TransferManager> or singleton anywhere to check first.
> 2. downloadFile()'s call site has no injection point. Compare it to the sibling uploadFileToAwsS3(S3TransferManager tm, ...) in the same file — that one does accept an optional pre-built tm ("if null, one will be temporarily created"), so upload already has the hook for a caller to pass in a shared client. downloadFile() has no such parameter; it hardcodes buildS3TransferManager(bro[1]) inline inside its own try (...) block. Reuse would need a signature change, not just a wrapper. 
> 3. The client is region-scoped, so "one" shared instance isn't quite right. region comes from parsing each object's own S3 URL per call. A real fix needs a small cache keyed by region (most ERDDAP deployments only ever hit one region, but the code doesn't assume that), not a single global client.
> 4. Lifecycle ownership has to move somewhere. Right now try-with-resources closes the client at the end of every single call — that's why it doesn't leak (post-#302) despite never being reused. A shared client must never be closed per-call; closing has to happen only on eviction or JVM/webapp shutdown instead. ERDDAP's SSR/File2 are static utility classes with no existing shutdown hook or lifecycle manager for this kind of resource, so that's new machinery, not a one-liner.
> 5. Credential refresh is a minor but real consideration. A client cached for the life of a long-running server needs its credentials provider to keep refreshing correctly (temporary AWS credentials rotate) — DefaultCredentialsProvider handles this internally, so it's probably fine, but it's the kind of thing you'd want to verify before caching indefinitely rather than assume.

> None of that is hard, but it's meaningfully more than the try (S3TransferManager tm = ...) one-liner that actually shipped for #302 — a thread-safe, region-keyed cache with real lifecycle management is a small design task, not a find-and-wrap fix. That's almost certainly why the original // !!! ??? reuse these??? comment stayed a comment instead of becoming code at the time.

Appreciate a look. Thanks so much

SethChampagneNRL:
# Description

Implement a cache for the S3TransferManagers to avoid memory leak from too many connections. 

Fixes #592 

## Type of change


ChrisJohnNOAA (approved):
I'm going to wait for the tests to complete and pass before merging, but the code itself looks good.

There was a test failure that I believe is flaky so I expect it to pass on the second run.

joe-smithe-glos:
@SethChampagneNRL Many thanks for the work. I haven't done Java since undergrad but the concepts stick and make sense as implemented here. 

Take care

CC @ChrisJohnNOAA

rmendels:
@joe-smithe-glos If you implement this please let Chris know if you run into any other problems.  As he noted this is an alpha-version of a future release with some new things that have not been extensively tested so you can count your self as our first alpha tester for this future release!  Thanks.

joe-smithe-glos:
Can do! Giving the RClone route a try first but will undoubtedly give `erddap/erddap:alpha-latest` a go in staging once the dust clears on the RClone attempt as the changes should help all around. Thanks!

